PR #9, opened 2016-04-14
-j1 -> RUST_TEST_THEADS=1

japaric:
the former limits the compilation phase to one rustc process, the later
executes the tests in a single thread. We actually want the later.


japaric:
@homu r+

PR #11, opened 2016-04-17
simplify rustup install

japaric:


japaric:
I want to use the sh.rustup.rs URL instead of my fork of the rustup.rs repo. So blocked on sh.rustup.rs getting updated.

PR #39, opened 2016-09-06
Add Debian build bug to readme

blutack:
Add a reference to the Debian CMake OpenSSL detection bug which breaks libssh2 builds. This reference can be removed once the package is moved from proposed-updates to updates.

Hopefully there is enough information in the description to at least point people in the right direction!


japaric:
> Add a reference to the Debian CMake OpenSSL detection bug which breaks libssh2 builds.

Oh :disappointed:.

Thank you for the heads up! :+1:

blutack:
Not your fault libssh2 is a bit of a beast! Copper & Xargo are really awesome!

PR #59, opened 2016-10-09
Added additional environment variables for building on OSX

regexident:
Building with just …

```bash
OPENSSL_ROOT_DIR=/usr/local/opt/openssl/ cargo install xargo
```

… fails on my machine (macOS 10.12).

It takes this mouthful …

```bash
LDFLAGS="-L/usr/local/opt/openssl/lib" \
CPPFLAGS="-I/usr/local/opt/openssl/include" \
PKG_CONFIG_PATH="/usr/local/opt/openssl/lib/pkgconfig" \
OPENSSL_ROOT_DIR=/usr/local/opt/openssl/ \
cargo install xargo
```

… to succeed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,8 +88,25 @@ If building on OSX 10.10 or greater, compilation of libssh2-sys will
 [fail due to missing OpenSSL headers](https://github.com/alexcrichton/ssh2-rs).
 Make sure OpenSSL is installed via HomeBrew, then run the following command:
 
+```bash
+OPENSSL_ROOT_DIR=$(brew --prefix openssl)/ \
+cargo install xargo
 ```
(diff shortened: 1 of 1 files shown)


japaric:
> It takes this mouthful …

Wow, thats awful.

Could you replace the hard-coded `/usr/local/opt/openssl` with `$(brew --prefix openssl)`. I think it produces the same thing; at least, that's why I use here in Travis CI.

Could you also change the `LDFLAGS` and `CPPFLAGS` to append the flags rather than override the existing ones?

At the end, it should probably should look like `LDFLAGS="$LDFLAGS -L$(brew --prefix openssl)/lib"`

regexident:
Done.

japaric:
Thank you!

regexident:
> Thank you!

No, thank _you_ for xargo (and copper)! 🙇

PR #89, opened 2016-11-19
Require sysroot for unsupported musl targets

dflemstr:
These musl targets aren't officially supported and thus don't have a pre-compiled `std` available for download.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,4 @@
-aaaaaa#![allow(dead_code)]
+#![allow(dead_code)]
 
 extern crate daggy;
 #[macro_use]
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,4 @@
-#![allow(dead_code)]
+aaaaaa#![allow(dead_code)]
 
 extern crate daggy;
 #[macro_use]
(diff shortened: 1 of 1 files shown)


dflemstr:
This might not be needed if the nightly build bots keep producing `std` versions for these targets, but I haven't been able to find the SLA for their availability.

japaric:
> This might not be needed if the nightly build bots keep producing std versions for these targets, but I haven't been able to find the SLA for their availability.

Huh? They are available via rustup:

```
$ rustup target add arm-unknown-linux-musleabi
```

Are you using some other way to install them?

Or, are they not available on same other channel like stable? Note that Xargo only works with a nightly Cargo/rustc.

dflemstr:
I guess I am getting confused because even though the targets are available via rustup there are linker errors, which made me think that they are not compiling against a stable toolchain.

See for example https://travis-ci.org/dflemstr/rq/jobs/177296948

Maybe it would be good to add a `--force-sysroot` flag to xargo instead, so that it builds one even if an official `std` is available, to avoid link problems?

dflemstr:
Will close this PR anyway

dflemstr:
(And I might be able to solve the above problem by doing some linker like [this](https://github.com/dflemstr/rq/blob/ae4eaf369f81ffb15c90a4a09e18755aad63a3c5/ci#L115-L140) so I'm not gonna open a new bug until I've verified that)

PR #92, opened 2016-12-14
treat custom toolchains the same as a nightly toolchain

oli-obk:


japaric:
I don't think this will work because xargo will look for the source in $sysroot/lib/rustlib/src/rust but won't find it there in the case of the "dev" channel.

homunkulus:
:umbrella: The latest upstream changes (presumably #94) made this pull request unmergeable. Please resolve the merge conflicts.

PR #101, opened 2016-12-26
Fix warning about SipHasher being deprecated.

thejpster:


japaric:
Cherry picked the top commit into PR #102 because this one has a few extra empty commits in it. I'm going to merge #102 instead of this PR.

Thanks for the PR, by the way!

PR #126, opened 2017-03-28
Add a note about a cargo issue with plugins (#125)

Thinkofname:
Not completely sure on the wording but I thought this was worth noting


japaric:
Thank you, @Thinkofname!

Issue #128, opened 2017-04-05
Xargo fails to copy nested directories


VictorKoenders:
After not using xargo for a couple of months I tried updating it today and I get the following error:

    D:\Development\Personal\rust_stm32f103>xargo --verbose
    + "rustc" "--print" "sysroot"
    + "rustc" "--print" "target-list"
    + "cargo" "build" "--release" "--manifest-path" "C:\\Users\\VKOEND~1.INT\\AppData\\Local\\Temp\\xargo.L9kWwUhdnTlo\\Cargo.toml" "--target" "thumbv7em-none-eabihf" "-v" "-p" "core"
    Compiling core v0.0.0 (file:///C:/Users/vkoenders.INTRANET/.rustup/toolchains/nightly-x86_64-pc-windows-msvc/lib/rustlib/src/rust/src/libcore)
     Running `rustc --crate-name core C:\Users\vkoenders.INTRANET\.rustup\toolchains\nightly-x86_64-pc-windows-msvc\lib\rustlib\src\rust\src\libcore\lib.rs --crate-type lib --emit=dep-info,link -C opt-level=3 -C metadata=ed3afdb7c1933857 -C extra-filename=-ed3afdb7c1933857 --out-dir C:\Users\VKOEND~1.INT\AppData\Local\Temp\xargo.L9kWwUhdnTlo\target\thumbv7em-none-eabihf\release\deps --target thumbv7em-none-eabihf -L dependency=C:\Users\VKOEND~1.INT\AppData\Local\Temp\xargo.L9kWwUhdnTlo\target\thumbv7em-none-eabihf\release\deps -L dependency=C:\Users\VKOEND~1.INT\AppData\Local\Temp\xargo.L9kWwUhdnTlo\target\release\deps -C link-arg=-Tstm32f103.ld -C link-arg=-nostartfiles --sysroot C:\Users\vkoenders.INTRANET\.xargo`
    Finished release [optimized] target(s) in 24.38 secs
    error: copying files from `C:\Users\vkoenders.INTRANET\.rustup\toolchains\nightly-x86_64-pc-windows-msvc\lib/rustlib\x86_64-pc-windows-msvc\lib` to `C:\Users\vkoenders.INTRANET\.xargo\lib/rustlib\x86_64-pc-windows-msvc\lib` failed
    caused by: Access is denied. (os error 5)
    stack backtrace:
     0:     0x7ff6ca37090f - <no info>
     1:     0x7ff6ca371222 - <no info>
[...]

PR #129, opened 2017-04-06
Fixed an issue where xargo could not copy files recursively on windows

VictorKoenders:
Fix for #128

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -11,9 +11,20 @@ pub fn cp_r(src: &Path, dst: &Path) -> Result<()> {
     (|| -> io::Result<()> {
             for e in src.read_dir()? {
                 let e = e?;
+                let metadata = e.path().metadata()?;
                 let f = e.file_name();
+                let src = src.join(&f);
+                let dst = dst.join(&f);
 
-                fs::copy(src.join(&f), dst.join(&f))?;
+                if metadata.is_dir() {
+                    fs::create_dir(&dst)?;
+                    cp_r(
+                        src.as_path(),
+                        dst.as_path()
+                    ).map_err(|e| io::Error::new(io::ErrorKind::Other, e.description()))?;
+                } else {
+                    fs::copy(src, dst)?;
+                }
             }
 
             Ok(())
--- a/src/util.rs
+++ b/src/util.rs
@@ -1,39 +1,44 @@
 use std::fs::File;
 use std::io::{Read, Write};
 use std::path::Path;
-use std::{fs, io};
+use std::fs;
 
 use toml::{Parser, Value};
+use walkdir::WalkDir;
 
 use errors::*;
 
 pub fn cp_r(src: &Path, dst: &Path) -> Result<()> {
-    (|| -> io::Result<()> {
-            for e in src.read_dir()? {
-                let e = e?;
-                let metadata = e.path().metadata()?;
-                let f = e.file_name();
-                let src = src.join(&f);
-                let dst = dst.join(&f);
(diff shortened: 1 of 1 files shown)


src/util.rs
```diff
@@ -11,9 +11,20 @@ pub fn cp_r(src: &Path, dst: &Path) -> Result<()> {
...
+                let src = src.join(&f);
+                let dst = dst.join(&f);
 
-                fs::copy(src.join(&f), dst.join(&f))?;
+                if metadata.is_dir() {
+                    fs::create_dir(&dst)?;
+                    cp_r(
+                        src.as_path(),
+                        dst.as_path()
+                    ).map_err(|e| io::Error::new(io::ErrorKind::Other, e.description()))?;
```
japaric: This is an interesting situtation! I think to avoid the back and forth between different error types you could write an "inner" function that does all the logic but always returns io::Result and then have cp_r call that inner function with chain_err. Something like this (not tested):

``` rust
pub fn cp_r(src: &Path, dst: &Path) -> Result<()> {
    fn cp_r_(src: &Path, dst: &Path) -> io::Result<()> {
        for e in src.read_dir()? {
            let e = e?;
            let metadata = e.path().metadata()?;
            let f = e.file_name();
            let src = src.join(&f);
            let dst = dst.join(&f);

            if metadata.is_dir() {
                fs::create_dir(&dst)?;
                cp_r_(src.as_path(), dst.as_path())?;
            } else {
                fs::copy(src, dst)?;
            }
        }

        Ok(())
    }

    cp_r_(src, dst).chain_err(
        || {
            format!(
                "copying files from `{}` to `{}` failed",
                src.display(),
                dst.display()
            )
        },
    )
}
```

VictorKoenders:
Actually I just saw we have a dependency  on walkdir 1.0.3 anyway, we might as well use that.
Would we be interested in logging the file that we error'd on and the original error?

japaric:
> Would we be interested in logging the file that we error'd on and the original error?

Yeah, it'd be great to have better diagnostics! 👍

src/util.rs
```diff
@@ -1,28 +1,44 @@
...
 
 use errors::*;
 
 pub fn cp_r(src: &Path, dst: &Path) -> Result<()> {
-    (|| -> io::Result<()> {
-            for e in src.read_dir()? {
-                let e = e?;
-                let f = e.file_name();
+    for e in WalkDir::new(src) {
+        let e = e.unwrap();
```
japaric: Hmm, shouldn't this error be lifted? I'm not sure about the exact failure mode here. Maybe it happens if the directory gets removed while walking it?

src/util.rs
```diff
@@ -1,28 +1,44 @@
...
+        let src_file = e.path();
+        let relative_path = src_file.strip_prefix(src).unwrap();
```
japaric: OTOH, this unwrap seems fine.

japaric:
Looks great now. 👍 

@homunkulus r+

homunkulus:
:pushpin: Commit eae6ab0 has been approved by `japaric`

homunkulus:
:hourglass: Testing commit eae6ab0 with merge eae6ab0...

homunkulus:
:sunny: Test successful - [status-appveyor](https://ci.appveyor.com/project/japaric/xargo/build/1.0.176), [status-travis](https://travis-ci.org/japaric/xargo/builds/219351933)
Approved by: japaric
Pushing eae6ab0749fdcc34a6b33aa48502963fe8b9d60b to master...

PR #143, opened 2017-05-25
README: automatically detect host platform target

RalfJung:
Add some shell magic so that the given command for using xargo with plugins actually works with all targets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,11 +161,14 @@ $ xargo build --target msp430-none-elf
     `rust/src/jemalloc` directory. `chmod -R +x rust/src/jemalloc` should do the
     trick.
 
-- When using compiler plugins (e.g. `serde_derive`) the target triple must
-  be provided even when compiling for the host platform due to the way
-  cargo handles compiler plugins. i.e. use `xargo build --target
-  x86_64-unknown-linux-gnu` instead of just `xargo build`. (Surprisingly, these
-  commands are NOT the same to Cargo)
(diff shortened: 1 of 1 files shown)


japaric:
I'd rather not make this change since that command won't work on Windows, and IMO it's clearer what the argument should be if the example shows a triple. rustc should have a `--print host` flag for this kind of thing though.

RalfJung:
`rustc -vV` does print the triple. That's what I am using here.

Maybe it would be useful to *also* have the more automatic one? I figured this out to be able to give instructions that work across architectures; others may also find it useful for the same reasons or e.g. for use in CI systems.

japaric:
In that case the text should say, with words not with code, that the host is in the output of `rustc -Vv`, and additionally show to how to extract the string on *nix.

RalfJung:
Done.

japaric:
Thanks @RalfJung!

japaric:
@homunkulus r+

homunkulus:
:pushpin: Commit 26305db has been approved by `japaric`

homunkulus:
:hourglass: Testing commit 26305db with merge 3353277...

homunkulus:
:sunny: Test successful - [status-appveyor](https://ci.appveyor.com/project/japaric/xargo/build/1.0.187), [status-travis](https://travis-ci.org/japaric/xargo/builds/236700817?utm_source=github_status&utm_medium=notification)
Approved by: japaric
Pushing 335327723b4883efbc5f5aae72559cf8a937fec8 to master...

PR #152, opened 2017-07-08
Pass `-C metadata` during sysroot builds

FenrirWolf:
This allows for Cargo/Xargo to properly distinguish between crates that are present both in the sysroot and included as `Cargo.toml` dependencies in downstream crates.

Not sure if it matters exactly what string is passed to the metadata argument or not, feel free to advise if it should be something better.

### DIFF
--- a/src/sysroot.rs
+++ b/src/sysroot.rs
@@ -73,6 +73,7 @@ version = "0.0.0"
             let mut cmd = Command::new("cargo");
             let mut flags = rustflags.for_xargo(home);
             flags.push_str(" -Z force-unstable-if-unmarked");
+            flags.push_str(" -C metadata=xargo");
             cmd.env("RUSTFLAGS", flags);
             cmd.env_remove("CARGO_TARGET_DIR");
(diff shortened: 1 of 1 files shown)


FenrirWolf:
After looking into things more, I found that Rustbuild also has to work around this issue, and it does so by having a shim for rustc that, among other things, intercepts and modifies any `-C metadata` arguments on-the-fly when building libs for the sysroot (see https://github.com/rust-lang/rust/blob/master/src/bootstrap/bin/rustc.rs#L41-L57 and https://github.com/rust-lang/rust/blob/master/src/bootstrap/lib.rs#L413-L429).

I'm not sure if the same thing could be accomplished purely in Xargo or if a rustc shim would also be necessary to apply a similar fix.

homunkulus:
:umbrella: The latest upstream changes (presumably #161) made this pull request unmergeable. Please resolve the merge conflicts.

FenrirWolf:
I found a better way around my issue after digging around some more, so Imma just close this.

PR #157, opened 2017-08-08
extend std caveats description

RalfJung:
I (think I) finally [understood](https://github.com/rust-lang/rust/issues/43637#issuecomment-320463578) why libstd behaves so strangely wrt. dependong on liballoc_jemalloc and libpanic_unwind. This extends the README to explain that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,8 +318,12 @@ stage = 2
   because `std` and other standard crates depend on unstable features so it's
   not possible to build the sysroot with stable or beta.
 
-- To build `std` *without* the "jemalloc" feature, apply the patch
-  in [rust-lang/rust#37975](https://github.com/rust-lang/rust/pull/37975).
+- `std` is always built as a dylib, which means it needs a panic library and an
+  allocator.  If you do not specify the `panic-unwind` feature, you have to set
+  `panic = "abort"` in `Cargo.toml`.  To build without the `jemalloc` feature,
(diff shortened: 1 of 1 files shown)


alexbool:
Great! Maybe you could also revive https://github.com/rust-lang/rust/pull/37975?

RalfJung:
With the new global allocators system, I am not sure if that's even going to work any more. Currently there is <https://github.com/rust-lang/rust/pull/43589> in flight which makes it so that building rustc with jemalloc disabled works at all. Once that lands, maybe we can have a way to control that from the outside when building just libstd. We *could* use the force_alloc_system flag, but that's explicitly not what it is intended for, and it would fix the allocator to be the system allocator for anything built against this libstd -- so, projects using xargo with a sysroot cold not make use of the global_allocators system. That doesn't seem desirable.

RalfJung:
Also notice that the changes done by <https://github.com/rust-lang/rust/pull/37975> (enabling the alloc_system feature and crate in more cases) are redundant nowadays; the feature and crate are actually unconditionally enabled.

The reference to that PR should be removed from the README, it is no longer useful and the patch no longer applies. One could make an equivalent PR for the new global allocator system, but that would work exactly like setting force_alloc_system: It would make it impossible for users of this libstd to pick their own allocator.

japaric:
Thanks @RalfJung.

PR #176, opened 2017-11-09
Search for Xargo.toml in parent directories.

Nils-TUD:
So far, xargo expected the Xargo.toml in the directory of the Cargo.toml. This patch searches for the Xargo.toml by walking up the directory hierarchy. This is useful if a workspace is used, leading, for example, to one Cargo.toml with the workspace and multiple Cargo.toml files for the members of the workspace. With this patch, it is no longer necessary to have one Xargo.toml for each Cargo.toml, but only one next to the workspace, for example.

### DIFF
--- a/src/xargo.rs
+++ b/src/xargo.rs
@@ -116,11 +116,10 @@ impl Toml {
 }
 
 pub fn toml(root: &Root) -> Result<Option<Toml>> {
-    let p = root.path().join("Xargo.toml");
-
-    if p.exists() {
-        util::parse(&p).map(|t| Some(Toml { table: t }))
-    } else {
+    if let Some(p) = util::search(root.path(), "Xargo.toml") {
+        util::parse(&p.join("Xargo.toml")).map(|t| Some(Toml { table: t }))
+    }
+    else {
         Ok(None)
     }
 }


RalfJung:
Sorry for the long silence. This seems useful! @jethrogb what do you think?

RalfJung:
Fair.

@Nils-TUD if you are still around, the patch looks reasonable -- but IMO it should come with a test. Could you add one?

Nils-TUD:
Actually, I switched to [cargo-xbuild](https://github.com/rust-osdev/cargo-xbuild), because Xargo didn't work for some time and cargo-xbuild works for me without changes  and doesn't even need Xargo.toml files.

Is there an advantage of Xargo over cargo-xbuild?

RalfJung:
> Actually, I switched to cargo-xbuild, because Xargo didn't work for some time and cargo-xbuild works for me without changes and doesn't even need Xargo.toml files.

Fair.

> Is there an advantage of Xargo over cargo-xbuild?

AFAIK, cargo-xbuild can only build libcore (maybe liballoc). Xargo can also build libstd, libtest, libproc_macro.

We can also close this PR if that's fine with you. Xargo is in low-effort maintenance mode, but I felt I should at least check the PRs that people opened forever ago.^^

Nils-TUD:
Ah, right, cargo-xbuild hardcoded the build of core, alloc and compiler_builtins. Then Xargo might indeed by interesting for me in the future :)

I'll come up with a test case.

Nils-TUD:
Sorry, I accidentally deleted the branch. Can you restore it?

But the good news is that I have a test :)

RalfJung:
I'm afraid you'll have to open a new PR. GH doesn't let me reopen the PR either.

Nils-TUD:
I think you just have reopen it, because I added a branch with the same name again.

RalfJung:
I can't though, GH won't enable that button.

Nils-TUD:
hm, okay, I'll open a new one.

PR #177, opened 2017-11-09
Added support for target-dir in config file.

Nils-TUD:
So far, xargo ignored the target-dir key in the config file. This caused the copy of the artifacts to the sysroot to fail if this key was used, because xargo expected them in the wrong place. This patch fixes this problem.

### DIFF
--- a/src/sysroot.rs
+++ b/src/sysroot.rs
@@ -34,6 +34,7 @@ fn build(
     ctoml: &cargo::Toml,
     home: &Home,
     rustflags: &Rustflags,
+    config: Option<&cargo::Config>,
     sysroot: &Sysroot,
     hash: u64,
     verbose: bool,
@@ -44,6 +45,11 @@ authors = ["The Rust Project Developers"]
 name = "sysroot"
 version = "0.0.0"
 "#;
+
+    let target_dir = match config {
+        Some(c) => c.target_dir()?,
+        None    => None,
+    };
 
     let rustlib = home.lock_rw(cmode.triple())?;
     rustlib
@@ -153,12 +159,13 @@ version = "0.0.0"
         }
 
         // Copy artifacts to Xargo sysroot
+        let base_dir = match target_dir {
+            Some(dir) => config.unwrap().root().join(dir),
+            None      => td.join("target"),
+        };
         util::cp_r(
-            &td.join("target")
-                .join(cmode.triple())
-                .join(profile())
-                .join("deps"),
-            &dst,
+            &base_dir.join(cmode.triple()).join(profile()).join("deps"),
+            &dst
         )?;
     }
 
@@ -221,6 +228,7 @@ pub fn update(
     root: &Root,
     rustflags: &Rustflags,
     meta: &VersionMeta,
+    config: Option<&cargo::Config>,
     src: &Src,
     sysroot: &Sysroot,
     verbose: bool,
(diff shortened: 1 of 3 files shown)


homunkulus:
:umbrella: The latest upstream changes (presumably #190) made this pull request unmergeable. Please resolve the merge conflicts.

RalfJung:
This looks reasonable (and sorry that it took so long to get you any feedback). Are you still interested in landing this?

src/sysroot.rs
```diff
@@ -112,13 +118,16 @@ version = "0.0.0"
...
-        )?;
+        match target_dir {
+            Some(dir) => util::cp_r(
+                &config.unwrap().root().join(dir).join(cmode.triple()).join(profile()).join("deps"),
+                &dst
+            ),
+            None      => util::cp_r(
+                &td.join("target").join(cmode.triple()).join(profile()).join("deps"),
+                &dst
+            ),
```
RalfJung: There's a lot of shared code between the two branches here, could you factor that out?

Nils-TUD:
Yes, I will rebase it onto the current master.

src/sysroot.rs
Nils-TUD: That's true, I'll improve that.

src/sysroot.rs
```diff
@@ -153,12 +159,13 @@ version = "0.0.0"
         }
 
         // Copy artifacts to Xargo sysroot
+        let base_dir = match target_dir {
+            Some(dir) => config.unwrap().root().join(dir),
```
RalfJung: If you fold the two matches into one here, the code becomes shorter and you avoid an `unwrap`.

Nils-TUD: No, because I want to do `td.join("target")` if `config` is None **or** `config.target_dir()?` is None. So the only way I can think of to avoid the unwrap is to have nested matches and have `td.join("target")` twice. Not sure if that's better.

RalfJung: Hm I see.
Either way please move `let target_dir` down to the other dir stuff here.

But then, wouldn't this work?
```rust
let base_dir = match config {
  Some(c) => c.target_dir()?.map(|d| c.root().join(d)),
  None => td.join("target"),
};
```

That's equivalent to your code, right?
However, it makes me wonder if this is really what we want. This directory will be outside the tempdir, right? We'll be spilling files into the user's dir and they might get mixed up with other stuff? Shouldn't this be more like `td.join(target_dir)` or so?

Nils-TUD: Ah, good idea. Yes, that's equivalent and better, I'll use that.

From my point of view it's the desired behavior. Without setting the target-dir, the files are put into a temp directory. With the target-dir, the files are put into that directory, relative to the root directory of the project. So for example, you can set it to `"build/sysroot"` and then all files go into `"<yourproject>/build/sysroot"`.

RalfJung: What is target-dir usually used for? Isn't that to change `target` to something else? So as a user I would expect by projects stuff to be put there, but not the sysroot.

Nils-TUD: hm, I created the patch quite some time ago, but I remember that without target-dir, Xargo was putting all kinds of things in the usual target directory. I just tested it again and actually, using a temp directory for the sysroot and using `CARGO_TARGET_DIR` to store the project stuff at a different place, looks fine. 

Maybe Xargo behaves differently now or I'm missing something. However, I would suggest that we close the PR for now and I report back if there still is a good reason for target-dir.

RalfJung: All right. Thanks a lot for looking into this!

PR #184, opened 2017-12-10
Add section to show how to disable jemalloc

### LINKED ISSUE #150: README out of date
The README currently says:

>To build std without the "jemalloc" feature, apply the patch in rust-lang/rust#37975.

This is no longer accurate; the patch is not necessary. Since https://github.com/rust-lang/rust/commit/70d2372adaf48bb24f8b829417229149bd103e4a, `std = { features = ["force_alloc_system"] }` is sufficient to disable jemalloc.

johnthagen:
Closes #150 

cc @RalfJung

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,8 +320,19 @@ stage = 2
 
 - `std` is built as rlib *and* dylib. The dylib needs a panic library and an
   allocator.  If you do not specify the `panic-unwind` feature, you have to set
-  `panic = "abort"` in `Cargo.toml`.  Currently, it is not possible to build
-  without the `jemalloc` feature.  For some further information on this issue,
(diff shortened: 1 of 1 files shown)


japaric:
Thanks, @johnthagen.

RalfJung:
Hm, maybe the README should mention that this is abusing `force_alloc_system` and could stop working any time? Also see <https://github.com/rust-lang/rust/issues/43637#issuecomment-320226134>.

japaric:
Perhaps Xargo should print a warning whenever a Xargo.toml file is used? Basically, trying to build anything other than just vanilla `core` (and maybe vanilla `std`?) can break at any time due to some nightly change (crates may be renamed, the name of Cargo features may change or the features themselves may disappear, etc.).

PR #200, opened 2018-03-08
Copy bin directory to host sysroot

FenrirWolf:
Allows custom targets to use the copy of `lld` bundled with Rust.

cc https://github.com/rust-lang/rust/issues/48772

PR #201, opened 2018-03-13
Update Cargo.lock

Woyten:
Since #200, `Cargo.lock` is considered dirty after Xargo has been built. This is particularly annoying when using Xargo as a Git submodule such as in tock/libtock-rs#18.

This PR updates `Cargo.lock` s.t. it will no longer be considered dirty after building Xargo.


japaric:
Thanks @Woyten 

bors r+

PR #204, opened 2018-03-21
Allowing stable channel rustc in Rust's style

dingelish:
[rust-sgx-sdk](https://github.com/baidu/rust-sgx-sdk) author here. We are facing a problem when using xargo in production. In production, we want to use stable branch of rust to compile the sysroot to achieve stability. However, xargo doesn't allow us to do this. We found that Rust provides a switch [`RUSTC_BOOTSTRAP`](https://github.com/rust-lang/rust/blob/b99172311c640c33f70676df7f75a899a999711c/src/libsyntax/feature_gate.rs#L1953)  which enables unstable features and allows stable branch rustc compiles latest libstd. And this option is enabled by default during bootstrap stage of building rust. So we believe that `RUSTC_BOOTSTRAP` is an acceptable solution for xargo and it's a desired feature.

Currently we have to maintain a branch of xargo for it. We wish it could be merged to mainstream. It's really helpful to us. Thanks!

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -140,13 +140,21 @@ fn run() -> Result<ExitStatus> {
                 sysroot.src()?
             },
             Channel::Stable | Channel::Beta => {
-                writeln!(
-                    io::stderr(),
-                    "WARNING: the sysroot can't be built for the {:?} channel. \
-                     Switch to nightly.",
-                    meta.channel
-                ).ok();
-                return cargo::run(&args, verbose);
+                if env::var("RUSTC_BOOTSTRAP").is_ok() {
+                    if let Some(src) = rustc::Src::from_env() {
+                        src
+                    } else {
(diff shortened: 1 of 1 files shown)


japaric:
Hello @dingelish 

As you say RUSTC_BOOTSTRAP is used in the bootsrap process and it's meant to *only* be used there. No tool should rely on that because it's an implementation of the bootstrap process; that is RUSTC_BOOTSTRAP could disappear, or the mechanism could change at any time and that would cause your "stable" use of it to break.

For that reason I'm not going to merge this PR. If you want a stable way to build a sysroot I suggest you push for *stable* Xargo in Cargo functionality (cf. rust-lang/cargo#4959) and notify the Cargo team (cc @aturon), who's in charge of that feature, about your need for it.

Issue #194, opened 2018-01-30
Xargo error: can't find specification for target


roblabla:
After updating my rust toolchain to latest nightly, running `xargo build -v --target=aarch64-nintendo-horizon-newlib` gives errors out with `Error loading target specification: Could not find specification for target "aarch64-nintendo-horizon-newlib"`. Full log can be found [in this gist](https://gist.github.com/roblabla/865fe5d27e3e76de00e4cec6d3dab615)

Xargo version is 0.3.10.

So after ensuring it didn't happen on my old toolchain, I tried to cut down which nightly version broke it, and it seems to be nightly-2017-12-25. On `nightly-2017-12-24`, everything works fine, on `nightly-2017-12-25`, I get the error above.

roblabla:
I suppose this is related to this commit: https://github.com/japaric/xargo/commit/f5349aa3fca776546b64407fc4be59f041c53e2a. However, given I'm running version 0.3.10, maybe that fix is not enough ?

roblabla:
My bad. This is a duplicate of https://github.com/japaric/xargo/issues/186 and https://github.com/rust-lang/cargo/issues/4905.

roblabla:
Actually reopening. I think the documentation/readme should be fixed to mention that people should set `RUST_TARGET_PATH=\`pwd\``

japaric:
> that people should set RUST_TARGET_PATH=`pwd`

That's a Cargo breakage. Older Cargos don't have that problem. You are free to send a pull request to update the documentation.

PR #205, opened 2018-03-22
Support absolute target JSON paths

phil-opp:
Builds upon https://github.com/rust-lang/rust/pull/49019 and https://github.com/rust-lang/cargo/pull/5228 to provide a solution to https://github.com/rust-lang/cargo/issues/4905 and https://github.com/japaric/xargo/issues/194.

The idea is that we allow targets ending in `*.json` (they were forbidden before) and convert them to canonicalized paths. Thus, they just work when compiling dependencies without the need to set the `RUST_TARGET_PATH` environment variable.

This PR just removes the explicit prohibition of paths as target and extends `Target::new` to handle such target paths. This should not lead to any breakage, because target paths ending in `.json` were explicitly forbidden before.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -56,9 +56,18 @@ impl CompilationMode {
         Ok(())
     }
 
+    /// Returns the condensed target triple (removes any `.json` extension and path components).
     fn triple(&self) -> &str {
         match *self {
             CompilationMode::Cross(ref target) => target.triple(),
+            CompilationMode::Native(ref triple) => triple,
+        }
+    }
+
+    /// Returns the original target triple passed to xargo (perhaps with `.json` extension).
+    fn orig_triple(&self) -> &str {
+        match *self {
+            CompilationMode::Cross(ref target) => target.orig_triple(),
             CompilationMode::Native(ref triple) => triple,
         }
     }
@@ -151,12 +160,7 @@ fn run() -> Result<ExitStatus> {
         };
 
         let cmode = if let Some(triple) = args.target() {
-            if Path::new(triple).is_file() {
-                bail!(
-                    "Xargo doesn't support files as an argument to --target. \
-                     Use `--target foo` instead of `--target foo.json`."
-                )
-            } else if triple == meta.host {
+            if triple == meta.host {
                 Some(CompilationMode::Native(meta.host.clone()))
             } else {
                 Target::new(triple, &cd, verbose)?.map(CompilationMode::Cross)
--- a/src/rustc.rs
+++ b/src/rustc.rs
@@ -111,7 +111,7 @@ impl Sysroot {
 #[derive(Debug)]
 pub enum Target {
     Builtin { triple: String },
-    Custom { json: PathBuf, triple: String },
+    Custom { json: PathBuf, triple: String, orig_triple: String },
 }
 
 impl Target {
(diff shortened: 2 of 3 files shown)


phil-opp:
(Do not merge yet, as the rustc and cargo PRs might need adjustments.)

japaric:
`cargo build --target ../foo.json` doesn't work when using Cargo due to rust-lang/cargo#4905 so I'd rather not make that special case work for Xargo. In general, I don't want Xargo to provide features that Cargo doesn't other than just building the sysroot because (a) that makes it harder to land Xargo's functionality in Cargo (should these extra features land as well?), and because (b) that could make people stick to Xargo when Cargo gains Xargo's functionality.

phil-opp:
I just realized that the PR description is a bit high level. Sorry about that! Let me explain in more detail:

> cargo build --target ../foo.json doesn't work when using Cargo due to rust-lang/cargo#4905 so I'd rather not make that special case work for Xargo.

I'm trying to fix rust-lang/cargo#4905 and already got a [cargo PR for this](https://github.com/rust-lang/cargo/pull/5228) merged. Target paths ending in `.json` work now from the cargo site and we even have a [test case](https://github.com/rust-lang/cargo/blob/715fc78caa832223a3404925a829e18c7832703a/tests/testsuite/custom_target.rs#L71) for `--target src/../custom-target.json`.

There is one remaining issue in `rustc` (see https://github.com/rust-lang/cargo/issues/4905#issuecomment-355947546), which will be fixed by https://github.com/rust-lang/rust/pull/49019. Afterwards, `rustc` and `cargo` will support target paths (ending in `.json`) and even convert them to absolute paths automatically, so that `cargo build --target foo.json` will work without `RUST_TARGET_PATH`.

However, when we try `xargo` with the new `rustc`/`cargo` versions, two issues arise:

1. `xargo` does an extra check to forbid targets ending in `.json`. 
2. `xargo` does not pass the original `--target` argument when building the sysroot, but only the triple without extension. Thus, the old `cargo` code path is executed, so that the sysroot metadata does not match to the desired target path, since target "foo" is a different target than a canonicalized "/home/…/foo.json" path).

I pushed a new version that makes it clearer that only the above two things are fixed.

phil-opp:
https://github.com/rust-lang/rust/pull/49019 was just merged and will be in the next nightly.

phil-opp:
The current nightly contains all necessary rustc and cargo changes, so this PR can now be tested via:

```
cargo install --git https://github.com/phil-opp/xargo.git --branch target-spec --force
```

Then `xargo build --target foo.json` (and also `--target src/../foo.json`) should work.

phil-opp:
@japaric Did you have some time to look at this? This PR is the only missing piece to allow compilation without `RUST_TARGET_PATH` again. It does not add any new features to `xargo`, it only removes the explicit prohibition of targets ending in `.json` and no longer modifies the target name that is passed to `cargo` (see https://github.com/japaric/xargo/pull/205#issuecomment-376230680 for details).

phil-opp:
Given that this PR has been open for over a year now without any progress, I'm going to give it a close. For what it's worth, we use this patch in [`cargo-xbuild`](https://github.com/rust-osdev/cargo-xbuild) without problems.

Issue #206, opened 2018-03-24
Issues with spaces in sysroot path (e.g., Windows user accounts containing spaces)


AaronCoad:
When using xargo build, the error "Multiple input filenames provided" is returned when the Windows user account contains a space as the folderpath isn't wrapped in double-quotes when passed to rustc --sysroot. In my case C:\Users\Aaron Coad\.xargo

On another Windows PC where there is no space within the user account, xargo build works successfully.

AxelMontini:
With #221 this should be fixed (escaping using `%20` or `\ `)

RalfJung:
`RUSTFLAGS` does not support any kind of escaping, I don't think that would help.

What could help though is to write a `.cargo` file instead of using an environment variable.

ZeWaka:
This is still an issue.

RalfJung:
This is a fundamental limitation in cargo: https://github.com/rust-lang/cargo/issues/6139

There's not much we can do about this on the xargo side, I am afraid.

kupiakos:
So, it seems the solution is to switch `RUSTFLAGS` to `CARGO_ENCODED_RUSTFLAGS` when passing to `cargo`. Possibly switching based on the cargo version.

> If the problem is xargo setting `RUSTFLAGS` unconditionally then that can be fixed with the recently added feature of `CARGO_ENCODED_RUSTFLAGS`. This is Cargo's "fix" to the solution for xargo.

https://github.com/rust-lang/cargo/pull/10232#issuecomment-1009012141

PR #207, opened 2018-03-25
Wrap sysroot path in double quotes as it might contain spaces

phil-opp:
Fixes https://github.com/japaric/xargo/issues/206


japaric:
Thanks @phil-opp.

bors r+

phil-opp:
Seems like cargo/rustc escapes single and double quotes passed in RUSTFLAGS. So this approach doesn't work. I'm not sure how we can work around this…

PR #213, opened 2018-04-08
compile core and compiler_builtins (+mem) ...

### LINKED ISSUE #212: Build compiler_builtins by default?
With https://github.com/rust-lang/rust/pull/49503 the compiler_builtins crate is injected to `no_std` crates by default. Thus, a `Xargo.toml` that builds `compiler_builtins` is always required now (see https://github.com/rust-lang/rust/pull/49503#issuecomment-377559413). Would it make sense to update `xargo` default to build `compiler_builtins` even without a `Xargo.toml`?

japaric:
if no Xargo.toml was specified

closes #212

cc @phil-opp

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ matrix:
     - env: TARGET=x86_64-apple-darwin
       os: osx
 
+    # Tests
     - env: TARGET=i686-unknown-linux-gnu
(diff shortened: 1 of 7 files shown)


japaric:
v0.3.12, which includes this change, has been published

PR #221, opened 2018-05-30
Escape sysroot and manifest paths to remove whitespaces

### LINKED ISSUE #206: Issues with spaces in sysroot path (e.g., Windows user accounts containing spaces)

AxelMontini:
## Hopefully fixes #206 

Fixes an error on windows where paths with whitespaces don't get escaped at all, causing the error `error: multiple input filenames provided`. In my case it happens because my Windows username is "Axel Montini" and xargo doesn't like the space in the middle.
This commit adds a function in utils to escape a string and replace spaces with `%20` on windows or `\ ` on other operative systems.
I haven't tested this on a Unix machine, but it seems to work on windows.

### DIFF
--- a/src/cargo.rs
+++ b/src/cargo.rs
@@ -45,7 +45,7 @@ impl Rustflags {
     pub fn for_xargo(&self, home: &Home) -> String {
         let mut flags = self.flags.clone();
         flags.push("--sysroot".to_owned());
-        flags.push(home.display().to_string());
+        flags.push(util::escape_argument_spaces(format!("{}", home.display())));
         flags.join(" ")
     }
 }
@@ -68,7 +68,7 @@ impl Rustdocflags {
     /// Stringifies these flags for Xargo consumption
     pub fn for_xargo(mut self, home: &Home) -> String {
         self.flags.push("--sysroot".to_owned());
-        self.flags.push(home.display().to_string());
+        self.flags.push(util::escape_argument_spaces(format!("{}", home.display())));
         self.flags.join(" ")
     }
 }
--- a/src/sysroot.rs
+++ b/src/sysroot.rs
@@ -129,7 +129,7 @@ version = "0.0.0"
                 }
             }
             cmd.arg("--manifest-path");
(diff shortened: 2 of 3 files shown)


phil-opp:
Hi @AxelMontini! Thank you for tackling #221!

It seems like `xargo` is no longer maintained, but we have an active fork at [rust-osdev/cargo-xbuild](https://github.com/rust-osdev/cargo-xbuild). Your PR looks useful to us, so I opened https://github.com/rust-osdev/cargo-xbuild/pull/6 as a copy of this PR. Is this ok with you that we merge it?

AxelMontini:
Hi @phil-opp! Yes, you can merge it if you want, so we can finally resolve this issue.

RalfJung:
These paths are used in `RUSTFLAGS`, right? Escaping won't help, unfortunately. Did you actually test this? I don't think this will work.

See https://github.com/rust-lang/cargo/issues/6139 -- with `RUSTFLAGS` there just is no way to pass spaces to `rustc`.

RalfJung:
Oh I see, I think this works on Windows because it actually replaces the space by a string that does not have a space. The unix version will definitely not work, there is still a space in the string and backslash does not help.

Can you update this PR to instead error on unix, pointing to https://github.com/rust-lang/cargo/issues/6139 ? We just cannot support paths with spaces on that platform currently. We can do that on Windows though, which seems nice!

phil-opp:
If I remember correctly, I removed this patch from `cargo-xbuild` again after it turned out that it wasn't really working.

RalfJung:
Okay, closing this PR then.
In principle I am open to a Windows-only solution to this problem, but then we should really have Windows CI as well to make sure this works.

PR #226, opened 2018-11-03
cargo update, recent rustc fixes, allow custom targets with embedded "."

evq:
This PR is related to the changes in https://github.com/rust-lang/cargo/pull/6255, originally discussed in https://github.com/rust-lang/rust/pull/55041#issuecomment-431275077

It allows for building using custom targets specs that have an embedded `.` with `xargo`.

I also had to make a few changes to get `xargo` to compile under the latest stable / nightly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,38 +1,78 @@
 [[package]]
-name = "backtrace"
-version = "0.3.0"
+name = "argon2rs"
+version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "backtrace-sys 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
- "cfg-if 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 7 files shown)


thejpster:
If you haven't seen, @phil-opp has forked Xargo to make https://github.com/rust-osdev/cargo-xbuild. That seems to be supported currently, while Xargo is in maintenance mode.

evq:
@thejpster there seems to be incompatibility between the way that `cargo xbuild` sets `TARGET`  and how `cortex-m-rt` expects it to be passed when using a json file target. `cargo xbuild` sets it to the full file path which breaks the prefix comparisons and thus linking with the bin files. I'm not sure if this is an issue that should be resolved in `cortex-m-rt` or `cargo-xbuild`.

RalfJung:
We have updated xargo for recent rustc separately and also just landed a `cargo update`, so I will close this PR.

Feel free to re-submit the part about embedded `.` -- but unfortunately the entire thing is just one commit so it is not clear which changes are relevant for that.

Issue #227, opened 2018-12-13
Broken with upcoming nightly


jethrogb:
Once a new nightly is produced, xargo will stop working because of https://github.com/rust-lang/rust/pull/56092 . You can already see this behavior when using https://github.com/kennytm/rustup-toolchain-install-master

```
$ xargo build --target x86_64-fortanix-unknown-sgx
   Compiling core v0.0.0 (/home/jethro/.rustup/toolchains/9fe5cb5342244a716055fa0162e795deabd4985c/lib/rustlib/src/rust/src/libcore)
    Finished release [optimized] target(s) in 22.12s
error: failed to load source for a dependency on `compiler_builtins`

Caused by:
  Unable to update /home/jethro/.rustup/toolchains/9fe5cb5342244a716055fa0162e795deabd4985c/lib/rustlib/src/rust/src/libcompiler_builtins

Caused by:
  failed to read `/home/jethro/.rustup/toolchains/9fe5cb5342244a716055fa0162e795deabd4985c/lib/rustlib/src/rust/src/libcompiler_builtins/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
error: `"cargo" "build" "--release" "--manifest-path" "/tmp/xargo.MmX2uglNXWpA/Cargo.toml" "--target" "x86_64-fortanix-unknown-sgx" "-p" "compiler_builtins"` failed with exit code: Some(101)
note: run with `RUST_BACKTRACE=1` for a backtrace
``

jethrogb:
> Once a new nightly is produced

Yay we finally got a new nightly :tada: Boo, xargo is now broken :disappointed:

thepowersgang:
What's the ETA on this landing in a release installable using `cargo install`

bruceloco:
Is this fixed in xargo 0.3.13?
I am trying for 3 days to build an arm-none-eabi for arm7 targets for kernel development and I am left scratching my head. First its create the target, its fun, but hard to find the info, then it fails because there are no binaries for it or libs even though you have the compiler(and this is in windows as well) then xargo shows as a suggestion and you move to nightly, and what not and the end result is this bug
If I wasnt a grown man I would cry :D

RalfJung:
Yeah this is long fixed. So if you have 0.3.13, whatever you are hitting is not this bug.

jethrogb:
[...]

PR #228, opened 2018-12-13
Fix for latest nightly: need to add a patch section

RalfJung:
Fixes https://github.com/japaric/xargo/issues/227

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,8 @@ matrix:
     # Tests
     - env: TARGET=x86_64-unknown-linux-gnu
       rust: nightly
+    - env: TARGET=x86_64-unknown-linux-gnu
+      rust: nightly-2018-12-01
 
 install:
   - bash ci/install.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,23 +5,23 @@ sudo: required
 
 matrix:
   include:
-    # Linux
-    - env: TARGET=i686-unknown-linux-gnu
-    - env: TARGET=i686-unknown-linux-musl
-    - env: TARGET=x86_64-unknown-linux-gnu
-    - env: TARGET=x86_64-unknown-linux-musl
-
-    # OSX
-    - env: TARGET=i686-apple-darwin
-      os: osx
-    - env: TARGET=x86_64-apple-darwin
-      os: osx
-
     # Tests
     - env: TARGET=x86_64-unknown-linux-gnu
       rust: nightly
     - env: TARGET=x86_64-unknown-linux-gnu
       rust: nightly-2018-12-01
+
+    # Linux (build only)
+    - env: TARGET=i686-unknown-linux-gnu
+    - env: TARGET=i686-unknown-linux-musl
+    - env: TARGET=x86_64-unknown-linux-gnu
+    - env: TARGET=x86_64-unknown-linux-musl
+
+    # OSX (build only)
+    - env: TARGET=i686-apple-darwin
+      os: osx
+    - env: TARGET=x86_64-apple-darwin
+      os: osx
 
 install:
   - bash ci/install.sh
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,38 +1,78 @@
 [[package]]
-name = "backtrace"
-version = "0.3.0"
+name = "argon2rs"
+version = "0.2.5"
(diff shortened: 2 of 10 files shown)


jethrogb:
Anyway, this (and #229) don't work (fully). It works if you're using only the `rust-src` component or `XARGO_RUST_SRC`. However, if you're specifying the source in `Xargo.toml`, the dependency on `rustc-std-workspace-core` will point to the wrong `libcore` source.

I've come up with 3 different approaches to fix this, not sure which one is best:

1. If there is a `std` crate in the blueprint, compute `../tools/rustc-std-workspace-core` from that path (or use the same git) and use that as the `[patch]`.

2. If there is a `std` crate in the blueprint, find the spec of the core dependency, then generate a `rustc-std-workspace-core` crate in the tempdir with that core (how would this work if `std` is a git dependency?). Then, use that temp crate as the `[patch]`.

3. Use the logic in this PR or #229 to generate the `[patch]` by default. Let users specify their own `[patch]`es in `Xargo.toml`

RalfJung:
I don't have a strong opinion about solving this problem.

However, I think we can merge one of #228 and #229 to fix the immediate problem for everyone using an unchanged libstd/`XARGO_RUST_SRC`, and then separately have an issue for fixing the other case?

jethrogb:
I think this PR doesn't work for compiling `core` without `Xargo.toml` so I'd vote #229

RalfJung:
> I think this PR doesn't work for compiling core without Xargo.toml so I'd vote

Not sure hat you mean by that, but fine for me.

However, please make sure xargo still works for old libstd that does not have `rustc-std-workspace-core` yet.

RalfJung:
Also I did the proper thing for fixing the deprecation warnings instead of just silencing them, feel free to copy those parts of my patches.

RalfJung:
Can't you check whether the rustc-std-workspace-core folder exists? That's what this patch does.

jethrogb:
How would that work for a git source? Xargo doesn't have access to the source.

RalfJung:
I'm afraid I am not familiar with the mode of operation of xargo you are referring to. I have only ever used xargo with the `rust-src` component, and I think it is reasonable to keep xargo compatible with older Rust nightlies -- it's not much effort.

japaric:
Thanks for the PRs, @RalfJung and @jethrogb! I'm going to land this one since it works with older nightly releases. On that note: @RalfJung, could you add a new builder that runs the test suite on an older nightly? I think the change should boil down to adding these lines to .travis.yml:

``` yaml
matrix:
  include:
    - env: TARGET=x86_64-unknown-linux-gnu
      rust: nightly-2018-12-XX
```

Feel free to land this with that change.

bors delegate+

japaric:
Hmm, I think you'll have to add `master` to the list of branches in .travis.yml.

I'll kick off a try build for now.

bors try

japaric:
Link to try build: https://travis-ci.org/japaric/xargo/builds/468425445

RalfJung:
Looks like CI is not in a good state...
```
E: Unable to locate package gcc-arm-none-eabi
```

RalfJung:
`ci/script.sh` does nothing if the branch is "master"... seems like this is not intended to work for PRs.  Odd.

bors try

RalfJung:
```
  failed to read `/home/travis/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libcompiler_builtins/Cargo.toml`
```
Well indeed, that file does not exist.

bjorn3:
@RalfJung `compiler_builtins` is one of those crates moved to crates.io.

RalfJung:
Ah then probably the testcases are wrong now.

RalfJung:
bors try

RalfJung:
Why is it not listening now?^^

bors try

RalfJung:
Oh it did listen, just GH didn't update. Never mind.

RalfJung:
Well that looks really green to me!

bors r+

RalfJung:
Uh, no, just the nightly-XXX didn't run the tests because it's not exactly "nightly".

bors r-

RalfJung:
bors try

RalfJung:
Okay, now it actually tested something and the tests come back green. :)

thepowersgang:
When would we expect to see this change in a released version on crates.io?

PR #234, opened 2018-12-19
Don't error on missing rustlib/$TRIPLE/bin

jethrogb:
```
error: intermittent IO error while iterating directory `.../rust/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/bin`
caused by: IO error for operation on .../rust/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/bin: No such file or directory (os error 2)
caused by: No such file or directory (os error 2)
note: run with `RUST_BACKTRACE=1` for a backtrace
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -417,7 +417,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "xargo"
-version = "0.3.12"
+version = "0.3.13"
 dependencies = [
  "dirs 1.0.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "error-chain 0.7.2 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/src/sysroot.rs
+++ b/src/sysroot.rs
@@ -276,16 +276,13 @@ pub fn update(
         &dst,
     )?;
 
-    let bin_dst = lock.parent().join("bin");
-    util::mkdir(&bin_dst)?;
-    util::cp_r(
-        &sysroot
-            .path()
-            .join("lib/rustlib")
-            .join(&meta.host)
-            .join("bin"),
-        &bin_dst,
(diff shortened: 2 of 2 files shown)


src/sysroot.rs
```diff
@@ -276,16 +276,12 @@ pub fn update(
...
-    util::cp_r(
-        &sysroot
-            .path()
-            .join("lib/rustlib")
-            .join(&meta.host)
-            .join("bin"),
-        &bin_dst,
-    )?;
+    let bin_src = sysroot.path().join("lib/rustlib").join(&meta.host).join("bin");
+    if bin_src.exists() {
```
RalfJung: Could you add a comment explaining why this is necessary, e.g., mention a `$TRIPLE` that has no `bin`? I am a bit confused why we need to copy this at all if `bin` is seemingly an optional directory.

jethrogb: I don't really remember. I just ran into this error once, noticed that I could create an empty directory that would make everything work, and filed this PR.

jethrogb: Looking on my filesystem, looks like this is where `lld` or `rust-lld` lives.

Cargo.lock
```diff
@@ -417,7 +417,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "xargo"
-version = "0.3.12"
+version = "0.3.13"
```
RalfJung: This diff seems surprising?

jethrogb: Surprising, but necessary. Forgotten in 590bdd23ff92fdc76b7e3fd57ddc6cd6785ce388

RalfJung: Ah, makes sense.

PR #241, opened 2019-04-26
 add patch for rustc-std-workspace-alloc 

RalfJung:
Blocked by https://github.com/japaric/xargo/pull/235 and https://github.com/rust-lang/rust/pull/60306.


src/sysroot.rs
```diff
@@ -307,17 +307,10 @@ impl Blueprint {
...
-            None => Table::new()
-        };
-
-        let rustc_std_workspace_core = src.path().join("tools/rustc-std-workspace-core");
-        if rustc_std_workspace_core.exists() {
+    /// Add $CRATE to `patch` section, as needed to build libstd.
+    fn add_patch(patch: &mut Table, mut path: PathBuf, crate_: &str) -> Result<()> {
+        path.push(crate_);
+        if path.exists() {
             // add rustc_std_workspace_core to patch section (if not specified)
```
jethrogb: This comment should be updated

jethrogb: This comment should be updated

RalfJung:
bors try

RalfJung:
bors try

RalfJung:
I *think* the reason for the failure is that a warning is printed due to the redundant `patch`, and the test is very sensitive to this kind of spurious output.

RalfJung:
bors try

RalfJung:
Yes, that did it!

PR #242, opened 2019-05-08
have CI in branches

RalfJung:
However, I am not entirely happy with this yes: (a) it builds all jobs, maybe we want to exclude some, and (b) won't this also run Travis on master after bors pushes staging to master? That seems silly.


RalfJung:
@jethrogb what do you think, does this look reasonable?

PR #246, opened 2019-08-03
let environment overwrite cargo command

RalfJung:
This is needed in https://github.com/rust-lang/rust/pull/63162 to make xargo work during rustc bootstrap.

r? @jethrogb

### DIFF
--- a/src/cargo.rs
+++ b/src/cargo.rs
@@ -142,8 +142,14 @@ fn flags(config: Option<&Config>, target: &str, tool: &str) -> Result<Vec<String
     }
 }
 
+pub fn command() -> Command {
+    env::var_os("CARGO")
+        .map(Command::new)
+        .unwrap_or_else(|| Command::new("cargo"))
+}
+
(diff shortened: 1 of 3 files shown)


RalfJung:
How would you test this? And do we have a test for `RUSTC` (where we already do the same)?

RalfJung:
Here's what we'd have to do, I think:
* find out the absolute path for rustc and cargo
* have a directory with dummy binaries for both that just always fail
* add that directory to the path, and set the env vars to the absolute paths we found earlier
* run xargo this way

TBH I don't think that's worth it...

RalfJung:
> r=me

bors=jethrogb

RalfJung:
bors r=jethrogb

jethrogb:
bors retry

RalfJung:
That did it. :) CI on the Rust side is passing.

@jethrogb could you make a new xargo release with this change, so that we can install from crates.io for Rust CI?

PR #254, opened 2019-09-09
 Search for Xargo.toml in parent directories, including test

Nils-TUD:
So far, xargo expected the Xargo.toml in the directory of the Cargo.toml. This patch searches for the Xargo.toml by walking up the directory hierarchy. This is useful if a workspace is used, leading, for example, to one Cargo.toml with the workspace and multiple Cargo.toml files for the members of the workspace. With this patch, it is no longer necessary to have one Xargo.toml for each Cargo.toml, but only one next to the workspace, for example.

### DIFF
--- a/tests/smoke.rs
+++ b/tests/smoke.rs
@@ -171,7 +171,13 @@ struct Project {
 }
 
 impl Project {
+    /// Creates a new project with given name in a temporary directory.
     fn new(name: &'static str) -> Result<Self> {
+        Self::new_in(std::env::temp_dir(), name)
+    }
+
+    /// Creates a new project with given name in a sub directory of `dir`.
+    fn new_in(dir: PathBuf, name: &'static str) -> Result<Self> {
         const JSON: &'static str = r#"
 {
     "arch": "arm",
@@ -186,7 +192,7 @@ impl Project {
 }
 "#;
 
-        let td = TempDir::new("xargo").chain_err(|| "couldn't create a temporary directory")?;
+        let td = TempDir::new_in(dir, "xargo").chain_err(|| "couldn't create a temporary directory")?;
 
         xargo()?
             .args(&["init", "--lib", "--vcs", "none", "--name", name])
@@ -343,7 +349,7 @@ fn simple() {
 }
 
 /// Test building a dependency specified as `target.{}.dependencies` in
-/// Xargo.toml
+/// ../Xargo.toml
 #[cfg(feature = "dev")]
 #[test]
 fn target_dependencies() {
@@ -351,8 +357,9 @@ fn target_dependencies() {
         // need this exact target name to get the right gcc flags
         const TARGET: &'static str = "thumbv7m-none-eabi";
 
-        let project = Project::new(TARGET)?;
-        project.xargo_toml(
+        let td = TempDir::new("xargo").chain_err(|| "couldn't create a temporary directory")?;
+        let project = Project::new_in(td.path().to_path_buf(), TARGET)?;
+        write(&td.path().join("Xargo.toml"), false,
             r#"
 [target.thumbv7m-none-eabi.dependencies.alloc]
 "#,
--- a/src/xargo.rs
+++ b/src/xargo.rs
@@ -121,11 +121,10 @@ impl Toml {
 }
 
 pub fn toml(root: &Root) -> Result<Option<Toml>> {
-    let p = root.path().join("Xargo.toml");
-
-    if p.exists() {
-        util::parse(&p).map(|t| Some(Toml { table: t }))
-    } else {
+    if let Some(p) = util::search(root.path(), "Xargo.toml") {
+        util::parse(&p.join("Xargo.toml")).map(|t| Some(Toml { table: t }))
+    }
+    else {
         Ok(None)
     }
 }


Nils-TUD:
The test "test" seems to be a bit unstable. It failed for me a couple of times as well, but now it works.

RalfJung:
That's a network failure indeed. I restarted that build.

RalfJung:
And now this fails due to https://github.com/rust-lang/rust/pull/64320. We'll have to wait until that lands.

RalfJung:
All right, this should work now then.

bors r+

RalfJung:
Looks like your test does not work with the 2018-12-01 nightly?

```
running: "cc" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "-mthumb" "-march=armv7-m" "-fno-builtin" "-fvisibility=hidden" "-ffreestanding" "-DVISIBILITY_HIDDEN" "-o" "/tmp/xargo.gIvG1WHQOidU/target/thumbv7m-parent-eabi/debug/build/compiler_builtins-588acbe8be9479e8/out/../../libcompiler_builtins/compiler-rt/lib/builtins/absvdi2.o" "-c" "../../libcompiler_builtins/compiler-rt/lib/builtins/absvdi2.c"

cargo:warning=cc: error: unrecognized command line option ‘-mthumb’
```

Is this a fairly recent target you are testing here? Could you use one used by other tests (and thus likely supported for longer already)?

Nils-TUD:
That's weird, because I did already copy the target from another test (and changed the middle part, as the other tests do, because apparently it's a problem if two have the same name). The `target_dependencies` test uses `thumbv7m-none-eabi`, which works according to the log. I'm not sure what's going on there...

RalfJung:
I am not sure either. I don't know anything about all this ARM/cross-compilation stuff.^^

Maybe it has something to do with the "none" target also occurring [here](https://github.com/japaric/xargo/blob/26494e7e7fcf0f349df768d1d53f39c97b5efb67/tests/smoke.rs#L184), whereas your target does not?

Nils-TUD:
Ok, it is interesting that the two other tests that have the comment "need this exact target name to get the right gcc flags" use an existing target (with "none"). So, maybe this is the problem.

As far as I can see, it should be an existing target, but cannot be a target that the host has already built. Otherwise it will find the alloc library in any case.

Let's try `riscv32i-unknown-none-elf` :)

RalfJung:
That does not seem to work...

> As far as I can see, it should be an existing target, but cannot be a target that the host has already built. Otherwise it will find the alloc library in any case.

Parallelism is another problem here -- tests are run in parallel. That's why the "host" tests have locking logic.

RalfJung:
What about modifying an existing test to have the toml file in a parent directory?

Nils-TUD:
> That does not seem to work...

But interestingly, my test (`target_dependencies_parentdir`) works. Only `host_patch` 
and `test` fail. And they fail, because `backtrace` is not found. I don't see why that is related.

> What about modifying an existing test to have the toml file in a parent directory?

That's exactly what I did, based on the `target_dependencies` test. But I can't use the same target, because this apparently causes those two tests to interfere with each other.

I'm a bit lost here. It works for me locally and with travis, only other and unrelated tests fail. Any suggestions?

RalfJung:
> But interestingly, my test (target_dependencies_parentdir) works. Only host_patch
and test fail. And they fail, because backtrace is not found. I don't see why that is related.

Ah, I should have checked the log then. ;)

> That's exactly what I did, based on the target_dependencies test. But I can't use the same target, because this apparently causes those two tests to interfere with each other.

No, what I meant is not to copy and modify, but to modify in-place.
But if what you got now works, that's even better.

tests/smoke.rs
```diff
@@ -172,6 +172,10 @@ struct Project {
 
 impl Project {
     fn new(name: &'static str) -> Result<Self> {
+        Self::new_in(std::env::temp_dir(), name)
+    }
+
+    fn new_in(dir: PathBuf, name: &'static str) -> Result<Self> {
```
RalfJung: So this will create a subdir and put everything into there, right? It won't work directly in `dir`?
Please add a doc comment explaining that. This had me confused for a second here.

Nils-TUD: Right, that is a bit unintuitive. I'll add a comment.

RalfJung:
Thanks a lot!

bors r+

Nils-TUD:
Why does it try to build the `arm/aeabi_cdcmp.S` when building for riscv?

hm, maybe I should really just modify the `target_depencencies` test in-place?

RalfJung:
I'm afraid I don't know.

We could change the minimal supported version to `nighty-2019-01-01` (mostly because that's a much less arbitrary date :D ), but I am not sure if making it one month younger will help with whatever issue you are running into here.

Nils-TUD:
I'm also fine with that, but I would suggest that I simply modify an existing test. I don't think that we lose anything with that.

RalfJung:
Yeah, this is fine. Thanks!

bors r+

RalfJung:
Yay, finally. :)

Nils-TUD:
Great :)

PR #256, opened 2019-09-14
clarify operating systems with available binaries

ayazhafiz:
This is very much a nit. Feel free to close and discard this PR if you wish.

It may also be worth getting rid of this line entirely. It appears that since 0.3.12 there have
only been Windows binary releases (excluding one linux one in 0.3.13).

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,8 +36,6 @@ panic=abort`, for your target.
 $ cargo install xargo
 ```
 
-But we also have [binary releases] for Linux, macOS, and Windows.
-
 [binary releases]: https://github.com/japaric/xargo/releases
 
 ## Usage
--- a/README.md
+++ b/README.md
@@ -36,8 +36,6 @@ panic=abort`, for your target.
 $ cargo install xargo
 ```
 
-[binary releases]: https://github.com/japaric/xargo/releases
-
 ## Usage
 
 ### `no_std`
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ panic=abort`, for your target.
 $ cargo install xargo
 ```
 
-But we also have [binary releases] for the three major OSes.
+But we also have [binary releases] for Linux, macOS, and Windows.
 
 [binary releases]: https://github.com/japaric/xargo/releases
 


RalfJung:
For 0.3.16, we seem to only have source releases on GH -- but when doing a release, I am just pushing a tag; then @japaric is doing "something" to add artifacts on GitHub (I don't even know how that works).

So probably it would make most sense to just remove the remark about binary releases, and point everyone at crates.io. Could you do that?

ayazhafiz:
Yep! I removed the remark entirely. Let me know id you'd rather the link the [crates.io page](https://crates.io/crates/xargo), though it doesn't seem to have anything not found in the README.

README.md
```diff
@@ -36,8 +36,6 @@ panic=abort`, for your target.
 $ cargo install xargo
 ```
 
-But we also have [binary releases] for the three major OSes.
-
 [binary releases]: https://github.com/japaric/xargo/releases
```
RalfJung: This is a dead Markdown reference now, is it not?

ayazhafiz: Fixed.

RalfJung:
Thanks!

bors r+

RalfJung:
Ah, this is blocked on getting https://github.com/rust-lang/rust/pull/64444 landed.

RalfJung:
bors retry

RalfJung:
bors retry

jethrogb:
> then @japaric is doing "something" to add artifacts on GitHub (I don't even know how that works).

Oh, huh, I thought this was automatic through CI, but it seems this didn't happen for 0.3.16...

RalfJung:
I see no indication in our CI that it would do anything like that. But I honestly don't know. I don't even know what it takes to turn a git tag into a GH "release"... not all tags seem to be treated as releases. Whatever.^^

jethrogb:
I was assuming [this](https://github.com/japaric/xargo/blob/master/.travis.yml#L43-L52) has something to do with it. If not, I don't understand what the `deploy` part does.

RalfJung:
Hm, yeah. No idea what that part does. And I hope that's not a secret API key we have in that file...

PR #257, opened 2019-09-19
Pass message-format to the cargo building the sysroot

roblabla:
Currently, `--message-format` is forwarded to the cargo building the final crate, but not when building the sysroot. This can break some tools that work by analyzing the output of `cargo build --message-format=json` when using xargo as a cargo replacement. (Split off from #217 )


RalfJung:
I have never used `--message-format` so I can't really comment on this, but the patch looks reasonable, so let's land it.

bors r+

PR #260, opened 2019-10-11
Make sure to use Xargo.toml directory when resolving path

### LINKED ISSUE #259: Canonicalize issue with Xargo.toml in parent directory
Since #254, xargo can search for Xargo.toml in parent directories.
However relative paths in the Xargo.toml aren't canonicalized using the correct root directory.

I made a repository to trigger the bug https://github.com/Thog/xargo_canonicalize_issue

This was tested on commit 08bf79ccf5d95ba1d5b71943145953e02eed0fa0

Thog:
This fix #259

### DIFF
--- a/src/sysroot.rs
+++ b/src/sysroot.rs
@@ -233,7 +233,7 @@ pub fn update(
     let ctoml = cargo::toml(root)?;
     let (xtoml_parent, xtoml) = xargo::toml(root)?;
 
-    let base_path = xtoml_parent.unwrap_or_else(|| root.path().to_path_buf());
+    let base_path: &Path = xtoml_parent.unwrap_or_else(|| root.path());
 
     let blueprint = Blueprint::from(xtoml.as_ref(), cmode.triple(), &base_path, &src)?;
 
--- a/src/sysroot.rs
+++ b/src/sysroot.rs
@@ -2,7 +2,7 @@ use std::collections::BTreeMap;
 use std::collections::hash_map::DefaultHasher;
 use std::hash::{Hash, Hasher};
 use std::io::{self, Write};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::{env, fs};
 
 use rustc_version::VersionMeta;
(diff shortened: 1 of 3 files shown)


RalfJung:
In https://github.com/japaric/xargo/issues/259 you talk about the path not being "canonicalized" correctly. I have no idea how that's an issue. Could you explain a bit more?

Is there any way the existing test for parent path Xargo.toml could be adjusted to cover this case (so that the bug does not come back)?

src/sysroot.rs
```diff
@@ -231,9 +231,11 @@ pub fn update(
     verbose: bool,
 ) -> Result<()> {
     let ctoml = cargo::toml(root)?;
-    let xtoml = xargo::toml(root)?;
+    let (xtoml_parent, xtoml) = xargo::toml(root)?;
 
-    let blueprint = Blueprint::from(xtoml.as_ref(), cmode.triple(), root, &src)?;
+    let base_path = xtoml_parent.unwrap_or_else(|| root.path().to_path_buf());
```
RalfJung: Seems a bit wasteful to make a copy here with `to_path_buf` if we only need an `&Path` below anyway. Can that be avoided?

jethrogb:
> In #259 you talk about the path not being "canonicalized" correctly. I have no idea how that's an issue. Could you explain a bit more?

@RalfJung When you use a relative path in `Xargo.toml`, you expect that to be resolved relative to the location of `Xargo.toml`. #259 claims that isn't done properly.

RalfJung:
Ah, yes that makes sense.

Is there any way we could have a test for this (e.g. adjusting the existing test)? (If that's too difficult we can proceed without, but I am worried that this could easily break again.)

RalfJung:
Thanks, the code looks good. Did you think about tests?

Also I think there should be a comment somewhere explaining what is going on. Clearly something is subtle here as the current code works in most but not all cases; any time the code does something subtle there should be comments pointing out that this is subtle and what it is that is being done.

RalfJung:
That's awesome. :) Let's see what CI says.

bors r+

PR #263, opened 2019-11-08
fix lockfile

RalfJung:
*oops* I forgot that in the last PR


RalfJung:
Self r+-ing because this just completes the other PR

bors r+

Cc @jethrogb

PR #264, opened 2019-11-08
CI: make sure the lockfile is up-to-date

RalfJung:
@jethrogb This seems like a useful thing to do on CI?

### DIFF
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -3,7 +3,7 @@ set -euxo pipefail
 beginswith() { case $2 in "$1"*) true;; *) false;; esac; }
 
 main() {
-    cross build --target $TARGET
+    cross build --target $TARGET --locked
     cross run --target $TARGET -- -V
 
     if beginswith nightly $TRAVIS_RUST_VERSION; then


RalfJung:
This is odd, CI should have failed because the lockfile actually isn't up-to-date...

jethrogb:
Are you sure `--locked` works properly when the only changes are not related to dependencies? There's also `--frozen` but I never know the difference.

RalfJung:
A `cargo build --locked` locally failed when I prepared this PR.

But maybe `cross` breaks `--locked`?

RalfJung:
Turns out that yes, [this is a `cross` issue](https://github.com/rust-embedded/cross/issues/346) and they are [working on it](https://github.com/rust-embedded/cross/pull/348).

Meanwhile, no reason not to land this here so that we get the fix once `cross` is fixed  -- @jethrogb what do you think?

RalfJung:
> There's also --frozen but I never know the difference.

`--frozen` implies `--locked`, but also refuses to download new crates into cargo's cache (it's basically "offline mode").

RalfJung:
The `cross` issue is fixed; I will land this here.

bors r+

PR #266, opened 2019-11-14
Set Cargo.lock writable

### LINKED ISSUE #265: 0.3.17 breaks when rust-src doesn't have u+w set
```
[leo60228@nixos:~/vloader]$ ls -Ll ~/.nix-profile/lib/rustlib/src/rust/Cargo.lock
-r--r--r-- 1 root root 212772 Dec 31  1969 /home/leo60228/.nix-profile/lib/rustlib/src/rust/Cargo.lock
```
`Cargo.lock` is copied with `std::fs::copy`, which copies permissions. This causes this error:
```
error: failed to write /tmp/xargo.qBhvt58W4S0y/Cargo.lock

Caused by:
  failed to open: /tmp/xargo.qBhvt58W4S0y/Cargo.lock

Caused by:
  Permission denied (os error 13)
```

leo60228:
Fixes #265

### DIFF
--- a/src/sysroot.rs
+++ b/src/sysroot.rs
@@ -102,7 +102,15 @@ version = "0.0.0"
 
         // rust-src comes with a lockfile for libstd. Use it.
         let lockfile = src.path().join("..").join("Cargo.lock");
-        fs::copy(lockfile, &td.join("Cargo.lock")).chain_err(|| "Cargo.lock file is missing from source dir")?;
+        let target_lockfile = td.join("Cargo.lock");
+        fs::copy(lockfile, &target_lockfile).chain_err(|| "Cargo.lock file is missing from source dir")?;
+
+        let mut perms = fs::metadata(&target_lockfile)
+            .chain_err(|| "Cargo.lock file is missing from target dir")?
+            .permissions();
+        perms.set_readonly(false);
+        fs::set_permissions(&target_lockfile, perms)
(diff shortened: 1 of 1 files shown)


jethrogb:
Thanks!

bors r+

src/sysroot.rs
```diff
@@ -104,6 +104,13 @@ version = "0.0.0"
         let lockfile = src.path().join("..").join("Cargo.lock");
         fs::copy(lockfile, &td.join("Cargo.lock")).chain_err(|| "Cargo.lock file is missing from source dir")?;
```
jethrogb: Can you save `td.join("Cargo.lock")` somewhere and reuse it?

src/sysroot.rs
```diff
@@ -102,7 +102,15 @@ version = "0.0.0"
 
         // rust-src comes with a lockfile for libstd. Use it.
         let lockfile = src.path().join("..").join("Cargo.lock");
-        fs::copy(lockfile, &td.join("Cargo.lock")).chain_err(|| "Cargo.lock file is missing from source dir")?;
+        let target_lockfile = td.join("Cargo.lock");
+        fs::copy(lockfile, &target_lockfile).chain_err(|| "Cargo.lock file is missing from source dir")?;
+
+        let mut perms = fs::metadata(&target_lockfile)
```
RalfJung: Would be good to have a comment here explaining what is happening any why.

PR #270, opened 2019-12-08
version 0.3.18

RalfJung:
r? @jethrogb

PR #272, opened 2019-12-31
version bump for 0.3.19 release

RalfJung:
I also used the opportunity for a `cargo update`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [v0.3.19] - ??
+
+### Added
+
+- Added a `xargo-check` binary that can be used for check-only builds of libstd
+  (thanks to @Aaron1011).
+
 ## [v0.3.18] - 2019-12-08
 
 ### Added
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [v0.3.19] - ??
```
jethrogb: Is this supposed to be a date?

RalfJung: I wasn't sure if we could get the release done today, so I didn't put in a date yet. I can make it 2019-31-12, but then we should make sure we get it through today. ;)

jethrogb: Let's put it this way, I'm not going to r+ without a date ;)

bors r+

RalfJung: You [did before](https://github.com/japaric/xargo/pull/270), but fair. ;)

jethrogb: I guess I was hoping for some CI that would fix it in post

RalfJung:
I'll back out the `cargo update`, that breaks testing with the old nightly... (maybe it's time to drop that test or so)

RalfJung:
bors r= jethrogb

PR #275, opened 2020-02-16
Use RUSTC_BOOTSTRAP to build sysroot for stable

### LINKED ISSUE #225: Do not require nightly
I've recently been told that setting the environment variable `RUSTC_BOOTSTRAP=1` makes a standard `rustc` build able to build nightly-only crates. Maybe this could be used to not require a nightly build for compiling libcore?

xurtis:
When building the sysroot or rust compiler with a beta or stable rustc
binary the flag RUSTC_BOOTSTRAP can be set to 1 to allow rustc to allow
the same features as the development channel. This is required as the
rust compiler and sysroot make use of unstable features.

See here: https://github.com/rust-lang/cargo/pull/5613

Fixes #225

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,10 +323,6 @@ which will perform a normal sysroot build, followed by a 'check' build of *your 
 
 ## Caveats / gotchas
 
-- Xargo won't build a sysroot when used with stable or beta Rust. This is
-  because `std` and other standard crates depend on unstable features so it's
-  not possible to build the sysroot with stable or beta.
-
 - `std` is built as rlib *and* dylib. The dylib needs a panic library and an
   allocator.  If you do not specify the `panic-unwind` feature, you have to set
   `panic = "abort"` in `Cargo.toml`.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -18,7 +18,7 @@ use std::path::{Path, PathBuf};
 use std::process::ExitStatus;
 use std::{env, io, process};
 
-use rustc_version::Channel;
+use rustc_version::Channel::*;
 
 use errors::*;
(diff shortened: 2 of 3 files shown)


RalfJung:
Thanks for the PR!

However, `RUSTC_BOOTSTRAP` is an internal rustc/cargo flag and not meant to be used by non-official projects. It can be used to subvert Rust's stability story.

I don't think we should do this.

xurtis:
The `RUSTC_BOOTSTRAP` flag is only used when building a sysroot, not when building user packages.

In general, building sysroot crates such as `core`, `alloc`, and `std` require setting the `RUSTC_BOOTSTRAP` flag to allow them to use stable compiler but with access to unstable features, without which these crates cannot be compiled.

If we wanted to avoid relying on this internal feature we'd probably have to wrap [`x.py`](https://github.com/rust-lang/rust/blob/master/src/bootstrap/README.md) instead (which would do this for us).

RalfJung:
Building a sysroot yourself is *not* something that is supported in a stable way by Rust. `RUSTC_BOOTSTRAP` is the internal flag rustc is using to build its own sysroot. To my knowledge, iIs not not ever, never meant to be used by outside projects.

xargo is used by people that build their own libstd; adding this flag would suddenly make that work on stable even though these custom libstd use tons of nightly features! That is violating Rust's safety story.

Let me put it this way: if you can get an official statement from the Rust Compiler team that it is okay for xargo to use this flag, then I am on board. But without such a statement, I'm afraid I will not accept this PR.

xurtis:
Alright, that seems reasonable.

PR #277, opened 2020-03-01
Fix custom std example in README

roblabla:
Compiler builtins is automatically built and linked as part of libcore - it is generally unnecessary to specify it in the Xargo.toml.

Adds a section detailing the (most common) cause of multiple matching crates error.

Replaces steed with rust-3ds as an example custom std facade, as it is an up-to-date, working example of a custom std facade.

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,10 +347,15 @@ which will perform a normal sysroot build, followed by a 'check' build of *your 
   host's target triple with `rustc -vV`. On *nix, the following rune will extract the triple:
   `rustc -vV | egrep '^host: ' | sed 's/^host: //'`.
 
-- Remember that `core`, `compiler_builtins` and `std` will get implicitly linked to your crate but
-  *all the other sysroot crates* will *not*. This means that if your Xargo.toml contains a crate
-  like `alloc` then you will have to add a `extern crate alloc` *somewhere* in your dependency graph
-  (either in your current crate or in some of its dependencies).
+- Remember that `core` and `std` will get implicitly linked to your crate but *all the other sysroot
+  crates* will *not*. This means that if your Xargo.toml contains a crate like `alloc` then you will
+  have to add a `extern crate alloc` *somewhere* in your dependency graph (either in your current
+  crate or in some of its dependencies).
+
+- Remember that rustc will always implicitly link `compiler_builtins` into your final binary, but
+  won't make it available for `use` the same way `core` and `std` are. So if you need to manually
+  call a `compiler_builtins` function, you will still need to manually add an
+  `extern crate compielr_builtins` within your crate.
 
 - Care must be taken not to end up with any "top-level" crates (`core`, `std`, `compiler-builtins`)
   twice in the sysroot. Doing so will cause cargo to error on build with a message like
--- a/README.md
+++ b/README.md
@@ -305,6 +305,19 @@ git = "https://github.com/rust3ds/ctru-rs"
 stage = 1
 ```
 
+### Patching sysroot crates
+
+Xargo also supports the `patch` feature from Cargo. This allows you to force the use
+of a custom crate throughout your sysroot's dependency tree. This can be especially
+useful to force the use of a custom `libc` or `compiler_builtins` without having to
+do intrusive changes to every transitive dependency.
+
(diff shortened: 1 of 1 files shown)


roblabla:
Pretty sure CI failure is not related to my README changes :eyes:

jethrogb:
bors try

RalfJung:
CI should be fixed by https://github.com/japaric/xargo/pull/278 -- but that bors job got cancelled for some reason?

README.md
```diff
@@ -351,11 +347,15 @@ which will perform a normal sysroot build, followed by a 'check' build of *your
   host's target triple with `rustc -vV`. On *nix, the following rune will extract the triple:
   `rustc -vV | egrep '^host: ' | sed 's/^host: //'`.
 
-- Remember that both `core` and `std` will get implicitly linked to your crate but *all the other
-  sysroot crates* will *not*. This means that if your Xargo.toml contains a crate like
-  `compiler_builtins` or `alloc` then you will have to add a `extern crate compiler_builtins` or
-  `extern crate alloc` *somewhere* in your dependency graph (either in your current crate or in some
-  of its dependencies).
+- Remember that `core`, `compiler_builtins` and `std` will get implicitly linked to your crate but
```
RalfJung: Is `compiler_builtins` really implicitly linked (by the prelude, or so)? So, I can write `compiler_builtins::stuff` without having written `extern crate compiler_builtins`?

README.md
```diff
@@ -288,25 +288,21 @@ sysroot. The final sysroot, the stage 1 sysroot, will contain both the `std` and
...
+[dependencies.alloc]
+[dependencies.panic_abort]
+[dependencies.panic_unwind]
 
 [dependencies.std]
-git = "https://github.com/japaric/steed"
-stage = 2
+git = "https://github.com/rust3ds/ctru-rs"
+stage = 1
 ```
```
RalfJung: One case that seems to come up regularly is something like
```
[patch.crates-io]
compiler_builtins = { ... }
```
Is that worth documenting here?

README.md
RalfJung: You wrote

> Compiler builtins is automatically built and linked as part of libcore - it is generally unnecessary to specify it in the Xargo.toml.

And indeed compiler_builtins is added to the crate through some magic (I thought it was a dependency of "core" but it is not). However, the comment you added here is not about specifying it in `Xargo.toml`, it is about specifying it in the project that uses this sysroot. That doesn't entirely fit together, does it?

I thought this was about which things one can actually `use` in a crate that is built with this sysroot, but that is yet another question.

README.md
roblabla: You're talking about patch in general, not specifically for compiler-builtins right?

Yeah, it is probably worth documenting somewhere. It can be useful to patch a handful of dependencies without doing intrusive changes (like using a custom libc, custom compiler_builtins, etc...).

README.md
roblabla: Eeh, the way I understand it, "linking" is about binary-level linking. I've never heard of people talking about making crates available for `use` as "linking".

It's true, however, that for basically every other crate, there's no difference between the two. Compiler-builtins is the only weird crate being linked implicitly at the binary level, but not implicitly available for `use` at the rust source level. I don't think anyone ever does things like calling a `compiler_builtins` function manually, but still, in the strictest sense, it *may* be necessary to do `extern crate compiler_builtins` to pull it within scope... Thanks I hate it.

I guess I'll move compiler_builtins to its own section to make it clearer.

README.md
RalfJung: Custom compiler_builtins seems like a common case based on the bugreports we get, but maybe that impression is wrong. Yeah, any example will do.

README.md
RalfJung: In hindsight I think I was mostly confused by you talking about `compiler_builtins` appearing in `Xargo.toml` in your PR summary, which seems unrelated to this linking discussion (the discussion doesn't talk about `Xargo.toml` at all, after all).

RalfJung:
Looking good, thanks! I have some more ideas for edits, will open a PR for that once this one lands.

bors r+

PR #280, opened 2020-03-03
edit README

RalfJung:
@roblabla does this look reasonable?

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,7 +321,7 @@ do intrusive changes to every transitive dependency.
 path = "path/to/custom/libc"
 ```
 
-Notice that you do should not list patched crates as `[dependencies]`!
+Notice that you should not list patched crates as `[dependencies]`!
 `[dependencies]` determines which crates are built in the first place; `[patch]`
 lets you replace some of their (transitive) dependencies with your own choice.
 Having a crate listed in both will likely lead to crate duplication.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -318,7 +321,12 @@ do intrusive changes to every transitive dependency.
 path = "path/to/custom/libc"
 ```
 
-## Check-only sysroot build
+Notice that you do should not list patched crates as `[dependencies]`!
```
roblabla:
```suggestion
Notice that you should not list patched crates as `[dependencies]`!
```

README.md
```diff
@@ -296,10 +296,13 @@ dependencies in the Xargo.toml file as you would do with Cargo.toml:
 [utest]: https://github.com/japaric/utest
 
 ``` toml
+# First build some standard crates.
 [dependencies.alloc]
 [dependencies.panic_abort]
 [dependencies.panic_unwind]
 
+# Then build our custom facade. This depends on the crates
```
roblabla: Instead of saying "This depends on the crates above", maybe instead say "This implicitly depends on the above crates being available in the sysroot".

RalfJung: Like so?

roblabla: :+1: looks good.

RalfJung:
Thanks. :)
@jethrogb this is ready now.

PR #283, opened 2020-03-25
remove AppVeyor config

RalfJung:
AppVeyor CI does not seem to be set up, so there's no point in having that config in the repository.


RalfJung:
Wait, what? AppVeyor never ran on any PRs in the last year, but now it is triggered here? What is even going on?^^

RalfJung:
I mean I guess the alternative would be to actually make AppVeyor work, and require it in bors. Not sure if I will have the time for that, though.

RalfJung:
Oh never mind, this is actually doing something -- it is building releases! Somehow that didn't work for 0.3.18 and 0.3.17, but now it seems to work again.

PR #288, opened 2020-04-12
Allow fallback on value package.rust-src if XARGO_RUST_SRC doesn't exist

jam1garner:
This allows for including a `package` table in Xargo.toml, which optionally can contain a `rust-src` string specifying a rust source path in lieu of the existence of the `XARGO_RUST_SRC` environment variable.

Example `Xargo.toml`:
```toml
[package]
rust-src = "/home/jam/a/dev/rust-skyline/rust-std-skyline/src"

[dependencies.std]
stage = 0
```

### DIFF
--- a/src/rustc.rs
+++ b/src/rustc.rs
@@ -108,6 +108,12 @@ impl Sysroot {
     }
 }
 
+impl From<PathBuf> for Src {
+    fn from(path: PathBuf) -> Self {
+        Self { path }
+    }
+}
+
 #[derive(Debug)]
 pub enum Target {
     Builtin { triple: String },
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -134,11 +134,17 @@ fn run(cargo_mode: XargoMode) -> Result<Option<ExitStatus>> {
         // We can't build sysroot with stable or beta due to unstable features
         let sysroot = rustc::sysroot(verbose)?;
         let src = match meta.channel {
-            Channel::Dev => rustc::Src::from_env().ok_or(
+            Channel::Dev => if let Some(env) = rustc::Src::from_env() {
+                Some(env)
+            } else {
+                xargo::toml_src(&root)?
+            }.ok_or(
                 "The XARGO_RUST_SRC env variable must be set and point to the \
                  Rust source directory when working with the 'dev' channel",
             )?,
             Channel::Nightly => if let Some(src) = rustc::Src::from_env() {
+                src
+            } else if let Some(src) = xargo::toml_src(&root)? {
                 src
             } else {
                 sysroot.src()?
--- a/src/xargo.rs
+++ b/src/xargo.rs
@@ -13,6 +13,7 @@ use errors::*;
 use extensions::CommandExt;
 use flock::{FileLock, Filesystem};
 use {cargo, util};
+use rustc::Src;
 
 pub fn run(
     args: &Args,
@@ -118,6 +119,11 @@ impl Toml {
     pub fn patch(&self) -> Option<&Value> {
         self.table.lookup("patch")
     }
+
+    /// Returns the `rust-src` part of Xargo.toml
+    pub fn package(&self) -> Option<&Value> {
+        self.table.lookup("package")
+    }
 }
 
 /// Returns the closest directory containing a 'Xargo.toml' and the parsed
@@ -129,4 +135,19 @@ pub fn toml(root: &Root) -> Result<(Option<&Path>, Option<Toml>)> {
     else {
         Ok((None, None))
     }
+}
+
+/// Returns the closest directory containing a 'Xargo.toml' and the parsed
+/// content of this 'Xargo.toml'
+pub fn toml_src(root: &Root) -> Result<Option<Src>> {
+    Ok(toml(root)?.1.map(|toml|{
+        toml.package().map(|package|{
+            if let Value::Table(table) = package {
+                let src = table.get("rust-src")?.as_str()?;
+                Some(Src::from(PathBuf::from(src)))
+            } else {
+                None
+            }
+        })
+    }).flatten().flatten())
 }


src/rustc.rs
```diff
@@ -108,6 +108,12 @@ impl Sysroot {
     }
 }
 
+impl From<PathBuf> for Src {
+    fn from(path: PathBuf) -> Self {
+        Self { path }
```
RalfJung, line 113: I am not sure if we really want to permit implicit conversion via `into` here -- can you make this an inherent `Self::new` instead?

RalfJung:
@jam1garner thanks for the PR!

I have to say however, that this feels rather ad-hoc. Could you talk a bit about why you want this new feature, and why the `[package]` section is the best one for this?

RalfJung:
The CI failure seems to also occur on master, but I haven't been able to reproduce it locally yet (still on it). It worked fine a few weeks ago when the last PR landed; might be a rustc regression.

jam1garner:
So my motivation for this is, admittedly, somewhat ad-hoc in nature. After having a lot of issues with git (primarily upstream issues from cargo and libgit2), I've found builds to be easier for users to reproduce through `XARGO_RUST_SRC`, but given I:

1. Don't want to further complicate the build process
2. Don't want to be dependent on something like `cargo-make`
3. Would rather not use a introduce a makefile or shell script or anything along those lines

I can't really set the Rust source path. Having a more permanent per-project means of configuring this felt appropriate.

As far as why the `[package]` section:
* Consistency with `Cargo.toml`
* Indicates that it is a package-level configuration
* Root-level string configurations in toml seem to not be idiomatic (at least from what I've seen from usage of tomls in the Rust community)
* Xargo's toml doesn't seem to have any top-level configuration yet (unless I'm missing something, I didn't do the deepest dive on the code) and having top-level configurations being "loose" feels both subjectively wrong and a bit short-sighted, as if any other such configurations are added they'd probably want to be grouped in some manner

I'm open to any suggestions as to how to better integrate something that serves these needs while fitting the existing structure of the project, but if this style of change is just not a good fit for Xargo I certainly don't mind if I have to continuously maintain this soft fork for my users.

Regardless, thanks for taking time to consider my PR!

RalfJung:
> After having a lot of issues with git (primarily upstream issues from cargo and libgit2), I've found builds to be easier for users to reproduce through XARGO_RUST_SRC

I am very confused... you are working around cargo issues by using xargo, but would not otherwise need xargo? Do you have a reference for where these upstream issues you mentioned are tracked? Seems better to fix them than to accumulate more hacks elsewhere in the ecosystem. ;)

Also your example `Xargo.toml` does not look very portable; what kind of setup do you envision where the same XARGO_RUST_SRC will work on many different machines? That would be my other concern -- typically the exact location where the sources reside is different from platform to platform and even from machine to machine, so having it in a version-controlled file seems counterproductive.

> Xargo's toml doesn't seem to have any top-level configuration yet

Yeah and that is exactly why I am wary of adding one, in this stage of xargo (maintenance-only, I am here mostly because I care about xargo working for the sake of [miri](https://github.com/rust-lang/miri/).)

jam1garner:
> I am very confused... you are working around cargo issues by using xargo, but would not otherwise need xargo?

I am using xargo for providing an stdlib implementation for an otherwise unsupported platform. I was mentioning my motivation for this PR, not for using xargo in general.

> Do you have a reference for where these upstream issues you mentioned are tracked?

Yes, not having shallow clones is actually the main issue that comes to mind -- rust-lang/cargo#1171

> Seems better to fix them than to accumulate more hacks elsewhere in the ecosystem. ;)

Y'know what? Fair. I'm afraid I don't quite have the time or motivation to try and make such large contributions to libgit2 (and then cargo) as would be required to push this along while the issue is still relevant to me. (Which, to be clear, is on me_

> Also your example Xargo.toml does not look very portable; what kind of setup do you envision where the same XARGO_RUST_SRC will work on many different machines? That would be my other concern -- typically the exact location where the sources reside is different from platform to platform and even from machine to machine, so having it in a version-controlled file seems counterproductive.

I mean yeah of course absolute paths aren't robust (admittedly the example I gave in the original comment was just the output of pwd that I had thrown into my toml for testing). What I actually use now is:

```toml
[package]
rust-src = "../rust-std-skyline-squashed/src"
```

The idea is that, for my use case, a user has a lot of projects linking against the same libstd and would like to ensure they are sharing the same copy (to save space and for control over updating). And, for other projects, including the rust source a submodule is reasonable.

RalfJung:
> I am using xargo for providing an stdlib implementation for an otherwise unsupported platform. I was mentioning my motivation for this PR, not for using xargo in general.

That is very useful information, thanks. Indeed it seems quite reasonable to not rely on xargo figuring out where the normal rust std source are as those are not very useful to you anyway. Suddenly it feels much less like a hack. ;)

Is `rust-std-skyline-squashed` in a git repo somewhere to look at, just to get a feel for it? You mention "including the rust source as a submodule"; I am not sure if you literally mean the full rustc repo with patches or something that just contains whatever libstd needs.

> The idea is that, for my use case, a user has a lot of projects linking against the same libstd and would like to ensure they are sharing the same copy (to save space and for control over updating). And, for other projects, including the rust source a submodule is reasonable.

Okay, so (for the cases where you don't use submodules) you are expecting the user to perform some setup steps to get a matching folder layout. Makes sense.

Honestly the submodule case sounds even more convincing to me, that could then even be entirely self-contained: `git clone`, submodule setup, `xargo build`, done.

jam1garner:
> Is rust-std-skyline-squashed in a git repo somewhere to look at, just to get a feel for it?

Yep! you can find it here: https://github.com/jam1garner/rust-std-skyline-squashed, it's a Github Actions squashed repo created from here: https://github.com/jam1garner/rust-std-skyline. Honestly it's nothing special, just a fork + squash of the rust-lang/rust repo.

> Okay, so (for the cases where you don't use submodules) you are expecting the user to perform some setup steps to get a matching folder layout.

Yep!

> Honestly the submodule case sounds even more convincing to me, that could then even be entirely self-contained: git clone, submodule setup, xargo build, done.

I certainly agree. If I didn't expect people to have a g good number of projects at once depending on my stdlib, I would absolutely have gone that route.

RalfJung:
> Yep! you can find it here: https://github.com/jam1garner/rust-std-skyline-squashed, it's a Github Actions squashed repo created from here: https://github.com/jam1garner/rust-std-skyline. Honestly it's nothing special, just a fork + squash of the rust-lang/rust repo.

Interesting.
So, assuming the cargo issue with shallow clones wouldn't exist, what would you do instead? You said this motivates having control over the source location in `Xargo.toml`, but to me it seems like even if you could use https://github.com/jam1garner/rust-std-skyline, you would have all the same problems?

src/cargo.rs
```diff
@@ -46,7 +46,7 @@ impl Rustflags {
     pub fn for_xargo(&self, home: &Home) -> String {
         let mut flags = self.flags.clone();
         flags.push("--sysroot".to_owned());
-        flags.push(home.display().to_string());
+        flags.push(format!("\"{}\"", home.display()));
```
RalfJung, line 49: Isn't this `RUSTFLAGS`? The quotes won't do anything there. There currently just is no way to pass arguments with spaces via `RUSTFLAGS`. That is a cargo bug: https://github.com/rust-lang/cargo/issues/6139

RalfJung:
Wow I had no idea there are xargo wrappers out there, let alone several of them... I hope they all have appropriate warnings that they rely on a tool that's basically unmaintained.^^

PR #291, opened 2020-05-03
make xargo fail on non-nightly, instead of continuing with the wrong sysroot

RalfJung:
This will likely just lead to failures later, so we better don't continue.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -144,13 +144,12 @@ fn run(cargo_mode: XargoMode) -> Result<Option<ExitStatus>> {
                 sysroot.src()?
             },
             Channel::Stable | Channel::Beta => {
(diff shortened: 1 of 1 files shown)


RalfJung:
@jethrogb any objections to this? Do you know of any way in which xargo on stable/beta is useful?

jethrogb:
This is fine with me.

> Do you know of any way in which xargo on stable/beta is useful?

Well I assume things “just work” when you set the bootstrap environment variable?

RalfJung:
> This is fine with me.

bors r+

> Well I assume things “just work” when you set the bootstrap environment variable?

Nobody should ever set that variable though... it's for rustc-internal use only.

PR #299, opened 2020-07-12
support relative XARGO_RUST_SRC

RalfJung:
We need to canonicalize the path because we change working directory before using it.


RalfJung:
Network failure.
bors retry

PR #307, opened 2020-12-07
add GitHub Actions-based CI

japaric:


japaric:
@RalfJung this PR ports the CI to GitHub Actions (GHA). The macOS build job is preserved. the `cargo_check_check_no_ctoml` test  fails with the 2019 toolchain but passes with a recently nightly toolchain. I can repro the test failure locally (with the 2019 toolchain; a 2020 toolchain works fine), even running the existing `ci/script.sh` (Travis CI commands) instead of the GHA commands. Any clue why that may be? 

Manually reproducing the `xargo-check` invocation from a Cargo project that doesn't contains a Cargo.toml is a no-op with the 2019 toolchain; the same invocation produces output with a 2020 toolchain.

RalfJung:
>  the cargo_check_check_no_ctoml test fails with the 2019 toolchain but passes with a recently nightly toolchain. I can repro the test failure locally (with the 2019 toolchain; a 2020 toolchain works fine), even running the existing ci/script.sh (Travis CI commands) instead of the GHA commands. Any clue why that may be?

No idea what's happening there.
How recent of a 2020 toolchain did you have to use to make it pass?

> Manually reproducing the xargo-check invocation from a Cargo project that doesn't contains a Cargo.toml is a no-op with the 2019 toolchain

Did you `rm ~/.xargo -rf`? The xargo test suite is rather fragile in how it depends on global state.

.github/bors.toml
```diff
@@ -0,0 +1,2 @@
+delete_merged_branches = true
+status = ["ci"]
```
RalfJung, line 2: What is this file about? I don't think I saw it before. There's also still `bors.toml` in the root.

japaric:
> How recent of a 2020 toolchain did you have to use to make it pass?

I only tried the nightly I had installed around: `2020-10-24`. I'll try again with the 2019 toolchain after removing the `~/.xargo` directory and see if that makes any difference.

.github/bors.toml
japaric: this is the same file as the `bors.toml` file that was in the root. the bors.toml can be either in the root or in the .github folder (apparently?). I can put back in the root to reduce churn.

japaric:
this appears to work:
``` console
$ rm -rf ~/.xargo
$ rustup default nightly-2019-12-01
$ # run only the test that was failing
$ cargo test --features dev -- ctoml
running 1 test
test cargo_check_check_no_ctoml ... ok
```

Now, let's see if I can make it fail again.

> Did you rm ~/.xargo -rf? The xargo test suite is rather fragile in how it depends on global state.

shouldn't CI do something like this (`rm -rf ~/.xargo`)? it runs tests in parallel and those test modify ~/.xargo (iirc)

japaric:
> Now, let's see if I can make it fail again.

running the test twice in a row fails:

``` console
$ rm -rf ~/.xargo
$ rustup default nightly-2019-12-01
$ # run only the test that was failing
$ cargo test --features dev -- ctoml
running 1 test
test cargo_check_check_no_ctoml ... ok

$ cargo test --features dev -- ctoml
running 1 test
test cargo_check_check_no_ctoml ... FAILED
```

`rm -rf ~/.xargo` makes it pass again

japaric:
> running the test twice in a row fails:

weird, this also fails with `nightly-2020-10-24`. maybe when tests are run in parallel this test has a non zero chance to pass / fail?

japaric:
OK, cleaning the *target* sysroot at the *start* of the `check_no_ctoml` fixes the issue locally. Let's see if GHA agrees.

I think the issue was that `check_no_ctoml` uses `HProject` but that doesn't clear the *target* sysroot. The test is cross compiling to windows so the target sysroot is only built if there wasn't one before in ~/xargo. All other tests that involve cross compilation use `Project` which clears the target sysroot when dropped.

japaric:
finally green. I think the issue in the CI was that I originally set ` CARGO_TERM_COLOR` to `always`; this caused `xargo-check` output to include ANSI escape codes for the colors and this made the `contains("Checking core")` assertion fail ("Checking" was colored and surrounded by ANSI escape codes)

japaric:
bors try
(let's see if bors works)

japaric:
bors try

japaric:
bors try

japaric:
bors try

japaric:
for some reason bors is no longer working on my existing repos (only on new ones) so I have changed branch protection to gate merging pull request on (GHA) CI passing.

RalfJung:
> shouldn't CI do something like this (rm -rf ~/.xargo)? it runs tests in parallel and those test modify ~/.xargo (iirc)

CI carefully (or accidentally) makes sure that all tests use different targets. Thus, if the `.xargo` folder is clean in the beginning, all tests will actually run. That was my interpretation of the behavior I saw, anyway.^^
Probably CI should use a per-test XARGO_HOME instead...

> for some reason bors is no longer working on my existing repos (only on new ones) so I have changed branch protection to gate merging pull request on (GHA) CI passing.

Eh, that's very strange...
https://app.bors.tech/repositories/5730 does see your PR though.

bors.toml
```diff
@@ -1,3 +1,2 @@
-status = [
-  "continuous-integration/travis-ci/push",
-]
\ No newline at end of file
+delete_merged_branches = true
```
RalfJung, line 1: Maybe this `delete_merged_branches` is causing trouble?

RalfJung:
Or maybe you ran into https://github.com/bors-ng/bors-ng/issues/806?

tests/smoke.rs
```diff
@@ -913,15 +913,18 @@ fn cargo_check_check() {
 #[test]
 fn cargo_check_check_no_ctoml() {
     fn run() -> Result<()> {
+        const TARGET: &str = "i686-pc-windows-gnu";
         let project = HProject::new(false)?;
+        // should be OK as long as other tests don't use the same target
```
RalfJung, line 918: Maybe add a comment that we need to do this because this is a cross-test using `HProject` instead of `Project`.

RalfJung:
bors ping

RalfJung:
@notriddle thanks a lot, seems to work again :)

tests/smoke.rs
japaric: done

japaric:
bors try
^ this should make "bors" show up in the "recent CI checks" section of the branch protection setting page. once it appears there I'll restore the old branch protection settings

japaric:
bors ping

japaric:
bors try

japaric:
> Already running a review

that "try" [failed](https://github.com/japaric/xargo/actions/runs/411422603) due to a syntax error so it didn't report a "ci" status; now bors thinks the "try" has not yet finished?

japaric:
bors try

japaric:
ok, updated the branch protection settings to gate on the "bors" status
bors r-

RalfJung:
I never really dug into the old CI setup, so I am happy for this to land if you consider it good enough. :)

PR #310, opened 2021-01-15
Fix copy failure by overriding CARGO_TARGET_DIR

### LINKED ISSUE #286: When a .cargo/config target-dir is set, xargo may fail
When a cargo target-dir is set, such as:

```
 > cat ~/.cargo/config
[build]
target-dir = "/Volumes/ramdisk/rs"
```

Xargo can fail, as the target directory is not where it expects, which can cause the cp_r utilities to fail unexpectedly. This was found in:  https://github.com/rust-lang/miri/issues/1285

oxalica:
Fixes #286 

According to @RalfJung (https://github.com/japaric/xargo/issues/286#issuecomment-607675112). It's better to use default `target`. So we just need to explicit override the target directory to fix the copy failure.

> xargo has its own workspace for building libstd. There is no interaction whatsoever with the workspace of your project. That's why I keep saying xargo should just overwrite the target dir.

### DIFF
--- a/src/sysroot.rs
+++ b/src/sysroot.rs
@@ -128,7 +128,10 @@ version = "0.0.0"
                 writeln!(io::stderr(), "+ RUSTFLAGS={:?}", flags).ok();
             }
             cmd.env("RUSTFLAGS", flags);
-            cmd.env_remove("CARGO_TARGET_DIR");
+
+            // Since we currently don't want to respect `.cargo/config` or `CARGO_TARGET_DIR`,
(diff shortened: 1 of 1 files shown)


RalfJung:
Thanks! Indeed I think this makes most sense. `td` is a temporary directory so this will not pollute people's working directories.

bors r+

RalfJung:
Eh, what the heck? The build looks all green...

RalfJung:
Let's try this again.
bors r+

Issue #215, opened 2018-04-10
Adjust expected Xargo.toml path if `--manifest-path` is passed


phil-opp:
When a `--manifest-path` argument is passed, a weird situation occurs: Cargo reads the project configuration from the passed `Cargo.toml` path, but Xargo reads the configuration from the current directory. This leads to errors if the project at the passed path needs additional crates such as `alloc`.

To fix this, we could update the `Xargo.toml` location when `--manifest-path` is passed. Would you be willing to merge a PR for this?

reuvenpo:
Is there a plan to implement this? I would like to be able to chanage the Xargo.toml path dynamically

RalfJung:
Xargo is in maintenance mode. So unless someone shows up and submits a PR, this is unlikely to be implemented.

Did you have a look at `cargo -Zbuild-std`, the intended successor for xargo? See https://github.com/rust-lang/wg-cargo-std-aware.

reuvenpo:
>unless someone shows up and submits a PR, this is unlikely to be implemented.

Ah, i see. fair enough, i suppose.

>Did you have a look at `cargo -Zbuild-std`

Unfortunately I'm using [an std that is maintained by a third party](https://github.com/apache/incubator-teaclave-sgx-sdk), but i can poke them to try and modernize their approach.

In any case, thanks for the quick reply!

jethrogb:
@reuvenpo have you tried using the [SGX target](https://edp.fortanix.com) that's included in the Rust compiler? No need to use Xargo.

reuvenpo:
Fortanix does not support compiling enclaves with strictly defined interfaces. We're building an enclave that is integrated in a larger application, implemented in a different language, and we want/need to use the enclave as a library. Part of our security model is having a very small attack surface on the enclave. 
([This has been a pretty successful approach so far](https://github.com/enigmampc/SecretNetwork), we just have a few small road bumps here and there which we've been working around.)

priyasiddharth:
@RalfJung can this issue be closed now?

RalfJung:
Ah yes, this was fixed by https://github.com/japaric/xargo/pull/312. Thanks :)

PR #312, opened 2021-03-24
fix: use manifest path for xargo.toml if provided

priyasiddharth:
Fix for https://github.com/japaric/xargo/issues/215

### DIFF
--- a/src/cargo.rs
+++ b/src/cargo.rs
@@ -250,14 +250,21 @@ impl Root {
     }
 }
 
-pub fn root(mode: XargoMode) -> Result<Option<Root>> {
+pub fn root(mode: XargoMode, manifest_path: Option<&str>) -> Result<Option<Root>> {
     // Don't require a 'Cargo.toml' to exist when 'xargo-check' is used
     let name = match mode {
         XargoMode::Build => "Cargo.toml",
         XargoMode::Check => "Xargo.toml"
     };
-    let cd = env::current_dir().chain_err(|| "couldn't get the current directory")?;
 
+    let cd = match manifest_path {
+        None => env::current_dir().chain_err(|| "couldn't get the current directory")?,
+        Some(p) => {
+            let mut pb = PathBuf::from(p);
+            pb.pop(); // strip filename, keep directory containing Cargo.toml
+            pb
+        }
+    };
     Ok(util::search(&cd, name).map(|p| Root { path: p.to_owned() }))
 }
 
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -6,7 +6,8 @@ pub struct Args {
     all: Vec<String>,
     subcommand: Option<Subcommand>,
     target: Option<String>,
-    message_format: Option<String>
+    message_format: Option<String>,
+    manifest_path: Option<String>,  // path to the Cargo toml file given in --manifest-path
 }
 
 impl Args {
@@ -35,19 +36,24 @@ impl Args {
     pub fn version(&self) -> bool {
         self.all.iter().any(|a| a == "--version" || a == "-V")
     }
+
+    pub fn manifest_path(&self) -> Option<&str> {
+        self.manifest_path.as_ref().map(|s| &**s)
+    }
 }
 
 pub fn args() -> Args {
     let all = env::args().skip(1).collect::<Vec<_>>();
 
-    let mut sc = None;
+    let mut subcommand = None;
     let mut target = None;
     let mut message_format = None;
+    let mut manifest_path = None;
     {
         let mut args = all.iter();
         while let Some(arg) = args.next() {
             if !arg.starts_with("-") {
-                sc = sc.or_else(|| Some(Subcommand::from(&**arg)));
+                subcommand = subcommand.or_else(|| Some(Subcommand::from(&**arg)));
             }
 
             if arg == "--target" {
@@ -58,14 +64,17 @@ pub fn args() -> Args {
                 message_format = args.next().map(|s| s.to_owned());
             } else if arg.starts_with("--message-format=") {
                 message_format = arg.splitn(2, '=').nth(1).map(|s| s.to_owned());
+            } else if arg.starts_with("--manifest-path") {
+                manifest_path = args.next().map(|s| s.to_owned());
             }
         }
     }
 
     Args {
-        all: all,
-        subcommand: sc,
-        target: target,
-        message_format: message_format
+        all,
+        subcommand,
+        target,
+        message_format,
+        manifest_path,
     }
 }
--- a/tests/smoke.rs
+++ b/tests/smoke.rs
@@ -216,7 +216,6 @@ impl Project {
 
         create_simple_project(td.path(), name, "#![no_std]")?;
         write(&td.path().join(format!("{}.json", name)), false, JSON)?;
-
         Ok(Project { name: name, td: td })
     }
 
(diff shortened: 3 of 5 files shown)


src/cargo.rs
```diff
@@ -250,14 +250,21 @@ impl Root {
...
         XargoMode::Build => "Cargo.toml",
         XargoMode::Check => "Xargo.toml"
     };
-    let cd = env::current_dir().chain_err(|| "couldn't get the current directory")?;
 
+    let cd = match path {
+        None => env::current_dir().chain_err(|| "couldn't get the current directory")?,
+        Some(p) => {
+            let mut pb = PathBuf::from(p);
+            pb.pop();
```
RalfJung, line 264: Why do you need to pop once?

RalfJung:
If possible it'd be great to also have a test for this. :) Our test suite is a bit fragile though, let me know if you run into trouble with it.

src/cargo.rs
priyasiddharth: pb is the path to `Cargo.toml`, popping gives the directory.

priyasiddharth:
> If possible it'd be great to also have a test for this. :) Our test suite is a bit fragile though, let me know if you run into trouble with it.

Will give it a go... thanks

src/cargo.rs
RalfJung: `pb` comes from `args.root()` which you documented to be the root *directory*... something seems off here?

src/cargo.rs
```diff
@@ -250,14 +250,21 @@ impl Root {
     }
 }
 
-pub fn root(mode: XargoMode) -> Result<Option<Root>> {
+pub fn root(mode: XargoMode, path: Option<&str>) -> Result<Option<Root>> {
```
RalfJung, line 253: Maybe call this  `root` or `root_dir` or so, for consistency with `Args`.

src/cli.rs
```diff
@@ -66,6 +74,7 @@ pub fn args() -> Args {
         all: all,
         subcommand: sc,
         target: target,
-        message_format: message_format
+        message_format: message_format,
+        root: root,
```
RalfJung, line 78:
```suggestion
        root,
```
Feel free to also do this for the other fields; this code likely predates that Rust feature.^^

src/cargo.rs
```diff
@@ -250,14 +250,21 @@ impl Root {
...
         XargoMode::Build => "Cargo.toml",
         XargoMode::Check => "Xargo.toml"
     };
-    let cd = env::current_dir().chain_err(|| "couldn't get the current directory")?;
 
+    let cd = match manifest_path {
+        None => env::current_dir().chain_err(|| "couldn't get the current directory")?,
+        Some(p) => {
+            let mut pb = PathBuf::from(p);
+            pb.pop();
```
RalfJung, line 264:
```suggestion
            pb.pop(); // strip filename, keep directory containing Cargo.toml
```

tests/smoke.rs
```diff
@@ -227,17 +227,30 @@ impl Project {
...
     /// Calls `xargo build` and collects STDERR
     fn build_and_get_stderr(&self, target: Option<&str>) -> Result<String> {
+        self.build_from_workdir_and_get_stderr(target, &self.td.path())
+    }
+
+    fn build_from_workdir_and_get_stderr(&self, target: Option<&str>, working_dir: &Path) -> Result<String> {
         let mut cmd = xargo()?;
         cmd.arg("build");
 
+        cmd.args(&["--manifest-path", &self.td.path().join("Cargo.toml").to_str().unwrap()]);
```
RalfJung, line 246: Doesn't this mean that now *all* tests run with `--manifest-path`?

tests/smoke.rs
```diff
@@ -445,6 +458,32 @@ fn dependencies() {
     run!()
 }
 
+/// Test building a dependency specified as `dependencies` in Xargo.toml
+#[cfg(feature = "dev")]
+#[test]
+fn dependencies_from_working_dir() {
+    fn run() -> Result<()> {
+        // need this exact target name to get the right gcc flags
+        const TARGET: &'static str = "thumbv6m-none-eabi";
```
RalfJung, line 467: Each test needs to have a different `TARGET`, to avoid using the same output directories. But the good part is that you can replace `none` by whatever you want and the targets still work -- see what the other tests are doing.

tests/smoke.rs
```diff
@@ -445,6 +458,32 @@ fn dependencies() {
     run!()
 }
 
+/// Test building a dependency specified as `dependencies` in Xargo.toml
```
RalfJung, line 461: I don't understand this comment. I thought this is about `--manifest-path`?

tests/smoke.rs
```diff
@@ -445,6 +458,32 @@ fn dependencies() {
...
+        project.xargo_toml(
+            r#"
+[dependencies.alloc]
+"#,
+        )?;
+        let td = TempDir::new("xargo_workdir").chain_err(|| "couldn't create a temporary directory")?;
+
+        project.build_from_workdir(TARGET, &td.path())?;
+        assert!(exists("core", TARGET)?);
+        assert!(exists("alloc", TARGET)?);
```
RalfJung, line 479: What does this have to do with `--manifest-path`?

tests/smoke.rs
priyasiddharth: Ack.

tests/smoke.rs
priyasiddharth: good catch! i'll fix this

tests/smoke.rs
priyasiddharth: AFAICT there are no targetted tests for cmd line args for me to emulate. I was thinking that if the project built successfully, it meant that the cmdline was being processed correctly. 

If there is an easier way - I would be interested to know.

RalfJung: Well, what happens in this test without the rest of your PR?

RalfJung: I always delete `~/.xargo` before re-running xargo tests since I think it will store some state there and you want to be sure to get a fresh run.

src/cargo.rs
priyasiddharth: fixed

src/cli.rs
priyasiddharth: fixed

tests/smoke.rs
priyasiddharth: fixed

priyasiddharth: fixed.

tests/smoke.rs
priyasiddharth: I've added  a `build_outside_project_dir` test which minimally tests the change.

tests/smoke.rs
priyasiddharth: fixed

tests/smoke.rs
```diff
@@ -445,6 +468,30 @@ fn dependencies() {
...
+        // need this exact target name to get the right gcc flags
+        const TARGET: &'static str = "thumbv6m-workdir-eabi";
+
+        let project = Project::new(TARGET)?;
+        project.xargo_toml(
+            r#"
+[dependencies.alloc]
+"#,
+        )?;
+        let td = TempDir::new("xargo_workdir").chain_err(|| "couldn't create a temporary directory")?;
```
RalfJung, line 485: So crucially, this is a `TempDir` that is *separate* from the `TempDir` inside `project`, right? If so, please add a comment explaining that.

priyasiddharth: correct.fixed.

RalfJung:
All right, looking good -- thanks @priyasiddharth!

bors r+

priyasiddharth:
> All right, looking good -- thanks @priyasiddharth!
> 
> bors r+

Thanks @RalfJung for the timely review!

PR #316, opened 2021-04-14
remove 'dev' feature

RalfJung:
I am not entirely sure why this was added, but not having `cargo test` run the main test suite has been eternally confusing... I think we better just get rid of this feature.


RalfJung:
Since there were no comments, I'll go ahead and merge this. We can always revert later.
bors r+

PR #317, opened 2021-04-15
release v0.3.23

RalfJung:
We should get these bugfixes shipped. @jethrogb any objections?


RalfJung:
I'll go ahead with the release now.
bors r+

PR #322, opened 2021-09-01
reference compiler-builtins by git source

### LINKED ISSUE #320: Cargo cannot build `compiler_builtins` for `aarch64-unknown-linux-musl`
_Full example on [this branch](https://github.com/littlebenlittle/wry-aarch64-musl-example/tree/xargo)_

I'm trying to build a [tauri-apps/wry](https://github.com/tauri-apps/wry) example for `aarch64-unknown-linux-musl`. My build container errors out, however.

The gist of the build is

```sh
ctr=`buildah from docker.io/library/rust:1.53-alpine3.13`
buildah run $ctr apk add \
	gtk+3.0 \
	webkit2gtk-dev \
	gtksourceview-dev
buildah add $ctr $LOCAL/aarch64-linux-musl-cross /opt/aarch64-linux-musl-cross
buildah run $ctr rustup target add aarch64-unknown-linux-musl
buildah run $ctr cargo install xargo
buildah run $ctr rustup default nightly
buildah run $ctr rustup component add rust-src
buildah commit $ctr webkit-aarch64
podman run -ti --rm \
	--name rs \
	-v ./:/mnt \
	-w /mnt \
	-e RUST_BACKTRACE=1 \
	localhost/webkit-aarch64 xargo build --target=aarch64-unknown-linux-musl --verbose

```

This results in

```
...
error: failed to parse manifest at `/tmp/xargo.yFwNkG10hwg9/Cargo.toml`

Caused by:
  dependency (compiler_builtins) specified without providing a local path, Git repository, or version to use.
error: `"cargo" "build" "--release" "--manifest-path" "/tmp/xargo.yFwNkG10hwg9/Cargo.toml" "--target" "aarch64-unknown-linux-musl" "-v" "-p" "compiler_builtins"` failed with exit code: Some(101)
...
```

~A challenge here is that the temporary Cargo.toml gets cleaned up xargo.~

> you can set XARGO_KEEP_TEMP to disable the cleanup -- @RalfJung

anickol:
Fixes #320, but I am not sure, what the previous behavior was!

### DIFF
--- a/src/sysroot.rs
+++ b/src/sysroot.rs
@@ -546,6 +546,11 @@ impl Blueprint {
                     "features".to_owned(),
                     Value::Array(vec![Value::String("mem".to_owned())]),
                 );
+                // reference compiler-builtins with `version = "*"`,
+                // the corresponding version of compiler_builtins matching the used std
+                // is selected because of the copied `Cargo.lock`-file from std.
+                cb.insert("version".to_owned(),
+                    Value::String("*".to_owned()));
                 cb.insert("stage".to_owned(), Value::Integer(1));
                 t.insert(
                     "compiler_builtins".to_owned(),


RalfJung:
> Is it because of the missing source for compiler_builtins in rust-src? I am not sure if it had been there before.

I don't think it has been there before, https://github.com/rust-lang/rust/pull/78209/files shows it has been pulled in from crates.io for a long time already.

But then why doesn't it pick up the version from https://crates.io/crates/compiler_builtins? The git reference will likely not match the exact version that this std should be built with.

RalfJung:
So the question is, in that default case where xargo adds a compiler_builtins dependency itself, which version should it add? Can we just set the version to `*` to get the latest version from crates.io? That seems less risky than using the latest git version...

anickol:
I will do some more research, but this should be the right path.

RalfJung:
To be specific, what do you mean by "this"? IMO using a crates.io version is preferable over using the git version.

anickol:
I tested if the version is somehow related to the src. 

When using `version = "*"` in the `Cargo.toml` the version of compiler_builtins is set to `0.1.50`. 

Without the version and without the git repository the version is set to `0.1.49`.

anickol:
Should we read the version of the dependency by parsing the Cargo.lock of the toolchain, e.g `.rustup/toolchains/nightly-x86_64-apple-darwin/lib/rustlib/src/rust/Cargo.lock`?

For my build it would be then: `compiler_builtins = { version = "0.1.49" }`

anickol:
I made a mistake. I think using `version = "*"' just works perfect, since we are copying the `Cargo.lock` from `std`, the correct version from `std` is selected! I will correct the checkin.

anickol:
On my machine the build even passes with the `nightly-2019-12-09`-toolchain when using the proposed solution.

PR #331, opened 2022-03-19
disable backtraces by default

RalfJung:
A significant fraction of the time for building xargo is spent on `backtrace` and its dependencies. In normal operation, backtraces shouldn't even come up, so I propose we disable this feature by default. Backtraces remain available as an optional feature (which can help with debugging).

(I care about xargo build times because `cargo miri setup` on CI is bottlenecked on building xargo.)

Also update dependencies while I am at it.


RalfJung:
@jethrogb any objections?
Otherwise I will go ahead and merge this and make a new release in a week or so.

jethrogb:
Seems fine

PR #343, opened 2022-09-15
replace unmainted `fs2` with `fs3`

dancrossnyc:
`xargo` would not build on illumos, due to a build failure in the unmaintained `fs2` crate.  Replace that with the `fs3` crate that has a current maintainer.

Signed-off-by: Dan Cross <cross@oxidecomputer.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -87,12 +87,13 @@ dependencies = [
 ]
 
 [[package]]
-name = "fs2"
-version = "0.4.3"
+name = "fs3"
+version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9564fc758e15025b46aa6643b1b77d047d1a56a1aea6e01002ac0c7026876213"
+checksum = "fb17cf6ed704f72485332f6ab65257460c4f9f3083934cf402bf9f5b3b600a90"
 dependencies = [
  "libc",
+ "rustc_version 0.2.3",
  "winapi",
 ]
 
@@ -133,9 +134,9 @@ checksum = "e2abad23fbc42b3700f2f279844dc832adb2b2eb069b2df918f455c4e18cc646"
 
 [[package]]
 name = "libc"
-version = "0.2.121"
+version = "0.2.132"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "efaa7b300f3b5fe8eb6bf21ce3895e1751d9665086af2d64b42f19701015ff4f"
+checksum = "8371e4e5341c3a96db127eb2465ac681ced4c433e01dd0e938adbef26ba93ba5"
 
 [[package]]
 name = "lock_api"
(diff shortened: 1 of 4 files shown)


RalfJung:
Thanks for the PR!

fs3 has way fewer recent users than fs2, and its [repo](https://github.com/oxidecomputer/fs3-rs) seems to be read-only, so it also looks unmaintained. (Also it seems to be created by you / your employer, which would have been good to mention.)

So I am not sure if this is strictly an improvement in terms of maintenance status?

dancrossnyc:
Thank you for the response back. I must be clear that I am not the author of fs3, but my understanding of it is that it was forked by a colleague because fs2 would not build on e.g. illumos (and indeed, it does not build there). I confess that I didn't even notice that it was under our corporate organization, but yes, that might have been good to mention.

As to this change itself, perhaps it's worth stepping back a moment and describing the overall objective, as the direction this PR goes in is really just an implementation detail.

I'd like to be able to use miri on an illumos host. While the current build has been broken there for a while now (note that I'm not complaining about that; illumos is a second-tier platform, miri is self-described as experimental, and we understand that and all the caveats that come with it), I had a version installed from an older nightly, and thought perhaps I could use it before trying to figure out how to get the most recent code building again. The first thing it wanted to do was install xargo, but the build failed because building fs2 failed, so that was my first step in debugging the whole thing. It turns out that the fs2 build failure was known, but the fs2 author is unresponsive to PRs, which was why fs3 was created. Put another way, `fs2` probably works well enough on Linux, macOS and perhaps Windows that this is mostly a non-issue for most users, but if one is on a more obscure platform, it doesn't work and there doesn't appear to be a way to get it fixed upstream, and patching locally is never fun.

Looking at xargo's use of fs2, it is only used for file-locking primitives, which e.g. cargo implements via wrapper calls to `libc`. There is an fs4, but it changes the locking API, and there are a few other dedicated flock-wrapper crates, but again, they all use different APIs. The easiest way to get xargo to build was to use fs3; hence this PR.

All that said, the objective is just to get xargo running on a second-tier platform, and if using `fs3` isn't the best way forward (even if it is perhaps the least effort) then I'd be happy to take a different tack: perhaps simply copying the relevant code from `cargo`?

What do you think?

RalfJung:
> I'd like to be able to use miri on an illumos host. 

Sounds good. :)  I won't be able to help with that directly, but am happy to facilitate what I can on the upstream side as long as it does not affect our major platforms.

Miri is moving away from xargo (https://github.com/rust-lang/miri/pull/2564), so if that is your goal then I'd suggest not to spend any time on xargo.

dancrossnyc:
> > I'd like to be able to use miri on an illumos host.
> 
> Sounds good. :) I won't be able to help with that directly, but am happy to facilitate what I can on the upstream side as long as it does not affect our major platforms.

Oh that's wonderful. Thank you!

> Miri is moving away from xargo ([rust-lang/miri#2564](https://github.com/rust-lang/miri/pull/2564)), so if that is your goal then I'd suggest not to spend any time on xargo.

That's fantastic news! It's not that I don't like xargo, but this sounds like an all-around improvement.

In that case, let's close this PR as effectively outdated.

If someone ever wants to get xargo specifically working on illumos, I'm sure they'll revisit, or perhaps the issue with `fs2` will have resolved itself by then.

PR #349, opened 2025-08-07
Fix building with Rust 1.89.0

svenpaulsen:
This includes the minimal necessary updates to get `xargo` compiling with the latest Rust toolchain. No functional changes were made - just adjustments to ensure compatibility... also applied code formatting using `cargo fmt`.

### DIFF
--- a/src/cargo.rs
+++ b/src/cargo.rs
@@ -3,13 +3,13 @@ use std::path::{Path, PathBuf};
 use std::process::{Command, ExitStatus};
 use std::{env, fmt};
 
-use toml::{Value, map::Map};
+use toml::{map::Map, Value};
 
 use cli::Args;
 use errors::*;
 use extensions::CommandExt;
-use util;
 use sysroot::XargoMode;
+use util;
 use xargo::Home;
 
 #[derive(Clone)]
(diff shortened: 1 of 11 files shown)


maratik123:
This fixes #350

bleggett (approved):
LGTM, also hit the same error and applied the same fix locally 👍

src/cargo.rs
```diff
@@ -50,7 +50,7 @@ impl Rustflags {
     pub fn encode(mut self, home: &Home) -> String {
         self.flags.push("--sysroot".to_owned());
         self.flags.push(home.display().to_string()); // FIXME: we shouldn't use display, we should keep the OsString
-        // As per CARGO_ENCODED_RUSTFLAGS docs, the separator is `0x1f`.
+                                                     // As per CARGO_ENCODED_RUSTFLAGS docs, the separator is `0x1f`.
         self.flags.join("\x1f")
```
neuschaefer, line 54: This formatting change looks semantically wrong. I think the first comment is about `home.display()...`, but the second one isn't a continuation of it, it's about `self.flags.join("\x1f")`. The way rustfmt will probably respect the old semantics is by having an empty line between them

